PR #76, opened 2017-10-24
Kctlread

SchwarzM:
Implements read commands into kubernikusctl to interact without a gui

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bin/
 _scratch
 _output
+*.swp
--- a/pkg/client/models/cluster_print.go
+++ b/pkg/client/models/cluster_print.go
@@ -1,0 +1,64 @@
+package models
+
+import (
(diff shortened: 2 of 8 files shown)


BugRoger:
rebase and quash please

SchwarzM:
done

.gitignore
```diff
@@ -1,3 +1,4 @@
 bin/
 _scratch
 _output
+*.swp
```
BugRoger: https://github.com/BugRoger/dotfiles/blob/master/vimrc#L47-L49 😅

BugRoger:
What was the problem with the `k.client == nil` issue the other day?

SchwarzM:
TBH i don't really know i moved the init into the PreRun of the cluster command and it worked.
I would still like it to be in PersistentPreRun of get cause all get commands will need it.

PR #81, opened 2017-10-27
Prevent deletion of kluster as long as nodepools are present

SchwarzM:
If nodes are still present during a terminate the routes will not be deleted from the neutron router.
As long as we do not implement a more sophisticated cleanup we should prevent klusters with 
nodepools from terminating.


pkg/api/handlers/terminate_cluster.go
```diff
@@ -20,7 +21,18 @@ type terminateCluster struct {
 
 func (d *terminateCluster) Handle(params operations.TerminateClusterParams, principal *models.Principal) middleware.Responder {
 
-	_, err := editCluster(d.Kubernikus.Kubernikus().Klusters(d.Namespace), principal, params.Name, func(kluster *v1.Kluster) {
+	k, err := d.Kubernikus.Kubernikus().Klusters(d.Namespace).Get(qualifiedName(params.Name, principal.Account), metav1.GetOptions{})
```
BugRoger: Let's extract this kluster a variable, so we can reuse it later...

pkg/api/handlers/terminate_cluster.go
```diff
@@ -20,7 +21,18 @@ type terminateCluster struct {
...
+		if apierrors.IsNotFound(err) {
+			return NewErrorResponse(&operations.TerminateClusterDefault{}, 404, "Not found")
+		}
+		return NewErrorResponse(&operations.TerminateClusterDefault{}, 500, err.Error())
+	}
+	if k.Status.NodePools != nil && len(k.Status.NodePools) > 0 {
+		return NewErrorResponse(&operations.TerminateClusterDefault{}, 409, "Cluster still has NodePools")
+	}
+
+	_, err = editCluster(d.Kubernikus.Kubernikus().Klusters(d.Namespace), principal, params.Name, func(kluster *v1.Kluster) {
```
BugRoger: Reuse kluster from above.

BugRoger: Also remove the double error handling below...

PR #90, opened 2017-11-03
Expose service and cluster CIDR in the api

databus23:
This allows users to optionally specify the service and cluster cidr when creating a cluster.
It can’t be changed afterwards.
By default it will still be set to

* service CIDR: `198.18.128.0/17`
* pod CIDR: `198.19.0.0/16`

We need this so we can re-create our parent clusters to avoid routing conflicts with the child clusters.

@SchwarzM Does `kubernikusctl` need changing to support creating a cluster with different CIDRs?


SchwarzM:
no i am using the marshal method of the generated client
as long as the generated client cluster spec has been updated (i can see the update)
you will have to provide a different cluster spec to the tool nothing else

databus23:
@BugRoger https://www.youtube.com/watch?v=yEEQ2FmWQQk#t=93

PR #96, opened 2017-11-13
introducing basic metrics

auhlig:
metrics:
- `kubernikus_kluster_status_phase{"kluster_id"="<id>","phase"="<phase>"} < 1|0 >`
- `kubernikus_node_pool_info{"kluster_id"="<id", "node_pool"="<name>", "image_name"="<name>", "flavor_name"="<name>"} <node_pool_size>`
- `kubernikus_node_pool_status{"kluster_id"="<id", "node_pool"="<name>", "status"="<status>"} < number of nodes in that status >`

Please add further metrics to this wishlist


pkg/metrics/exporter.go
```diff
@@ -0,0 +1,81 @@
...
+		ch <- metric
+	}
+	for _, metric := range e.nodePoolMetrics {
+		ch <- metric
+	}
+}
+
+func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
+
+  //TODO list klusters via API and derive the following metrics
```
auhlig: Do we want to go as described above or directly integrate in controllers?

BugRoger: We can put it in the controllers as whitebox metrics. Let's do it there instead of carrying around another binary that needs transport, supervision and stuff. And isn't even able to look into the applications.

I'd say we make it easy and cheap to put instrumentation all over the place. Complement and reduce the required logging...

auhlig:
Could you take another look @BugRoger?

databus23:
Much better. We definitely need to squash the two commits to avoid having all those files in the history.

auhlig:
Squashed. I specified the subpackages, so it's not pulling so much in the first place.

auhlig:
Better now @databus23? Any additional metrics?

auhlig:
Note to self: Use `phase` instead of `status`

PR #104, opened 2017-11-17
Nginx ingress

### LINKED ISSUE #80: Ingress Default Backend doesn't know JSON
Implement a default backend for the nginx-ingress that speaks html and json. This is required for various clients to be able to detect and report an error.

auhlig:
we want the ingress custom backend is already deployed in the admin cluster in the kubernikus control plane. 
official helm chart for nginx-ingress as base but with custom backend that also knows json. fixes #80.

Initially I wanted to use it as a dependency but override the image of the custom backend, but that doesn't seem to work.

### DIFF
--- a/charts/kubernikus-system/charts/nginx-ingress/.helmignore
+++ b/charts/kubernikus-system/charts/nginx-ingress/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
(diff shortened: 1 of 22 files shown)


charts/kubernikus-system/nginx-ingress/values.yaml
```diff
@@ -0,0 +1,245 @@
+## nginx configuration
+## Ref: https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md
+##
+controller:
+  name: controller
+  image:
+    repository: gcr.io/google_containers/nginx-ingress-controller
+    tag: "0.9.0-beta.14"
```
databus23: Any reason we are not using latest (beta.17)?

auhlig: Right. I'll test in staging first and update this PR afterwards

auhlig:
Squashed, rebased and manually deployed to `k-staging` cluster. The custom backend currently only handles 404, 500

auhlig:
Wanted to test the new controller in staging first, but it's a mess since they changed all nginx-based annotations :-/ Should be easier to get it in here.

PR #105, opened 2017-11-23
rename and add some more metrics. also add metric unit test

auhlig:


auhlig:
no discussion expected here. just wanted to make sure this builds.

PR #108, opened 2017-11-27
phase transition summary

auhlig:
Wanted something like [this](https://github.com/sapcc/kubernikus/compare/phasesum?expand=1#diff-9bc930f31cf0708f07b8d78355750f5cR47) @BugRoger?
`kubernikus_kluster_phase_duration{from_phase="Creating",to_phase="Running",quantile="0.99"} 10.00000009`

### DIFF
--- a/pkg/controller/ground.go
+++ b/pkg/controller/ground.go
@@ -192,6 +192,7 @@ func (op *GroundControl) handler(key string) error {
 				if err := op.updatePhase(kluster, models.KlusterPhaseRunning, ""); err != nil {
 					glog.Errorf("Failed to update status of kluster %s:%s", kluster.GetName(), err)
 				}
+				metrics.SetMetricBootDurationSummary(kluster.GetCreationTimestamp().Time, time.Now())
 				glog.Infof("Kluster %s is ready!", kluster.GetName())
 			}
 		case models.KlusterPhaseTerminating:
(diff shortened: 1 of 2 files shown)


BugRoger:
Yeah, cool! 👍

databus23:
Cool! But I'd like to point out that the calculation might not be accurate. The Kluster starts in phase `Pending` and is put in `Creating` by the operator. If the operator is not running that might take an arbitrary amount of time. So comparing with the creation time is not measuring the transition from one phase to the next but from initial creation of the resource to the  ready state.

auhlig:
True @databus23. It might make sense to add a timestamp when the phase was updated to the kluster metadata. Yes?
Else I'm not sure how to track this in the operator without screwing up the metric when the operator gets restarted.

auhlig:
After discussion with @databus23: changed to `kubernikus_kluster_boot_duration` to measure duration until kluster goes from pending to running. No new timestamp.

auhlig:
Note: Had to disable the [test](https://github.com/sapcc/kubernikus/pull/108/commits/c384fcd7fc7d9b2792a9c1d8f0bbc1d48b3fa1d0#diff-9bc930f31cf0708f07b8d78355750f5cR45) for the metric, as the value seems to differ in the last digit quite often. `kubernikus_kluster_boot_duration{quantile="0.99"} 10.00000003` vs `kubernikus_kluster_boot_duration{quantile="0.99"} 10.000000039`

auhlig:
rebased and squashed

PR #111, opened 2017-12-06
Masquerading in static firewall deployment

SchwarzM:
The static firewall currently allows ingress traffic on default ephemeral ports.
If masquerade has a port collision it will by default choose a port from 1024-32000 which is
outside the default ephemeral port range.

Thus the answer packet is never reaching our system.

We disable the non-masquerade-cidr which disables all masquerading and insert our own rules for masquerading which adheres to the correct port range.

### DIFF
--- a/pkg/templates/node.go
+++ b/pkg/templates/node.go
@@ -169,6 +169,7 @@ networkd:
 storage:
   files:
     - path: /var/lib/iptables/rules-save
+      filesystem: root
       mode: 0644
       contents:
         inline: |-
--- a/pkg/templates/node.go
+++ b/pkg/templates/node.go
@@ -15,6 +15,8 @@ locksmith:
 
 systemd:
   units:
+    - name: iptables-restore.service
+      enable: true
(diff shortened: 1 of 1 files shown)


pkg/templates/node.go
```diff
@@ -79,7 +81,7 @@ systemd:
           --cluster-dns={{ .ClusterDNSAddress }} \
           --cluster-domain={{ .ClusterDomain }} \
           --client-ca-file=/etc/kubernetes/certs/kubelet-clients-ca.pem \
-          --non-masquerade-cidr={{ .ClusterCIDR }} \
+          --non-masquerade-cidr=0.0.0.0/0 \
```
databus23: I think we can remove the setting completely. Should have the same effect.

databus23: I stand corrected. The default seems to be `10.0.0.0/8` which would not be what we want. Setting it to `0.0.0.0/0` seems to be the way to go.

PR #116, opened 2017-12-11
WIP: Refactor nanny/route gc to run in the operator

databus23:
Instead of deploying a separate binary for each kluster we decided to have this run centrally in the operator to make it easier to ship manage the lifecycle of the nanny components


databus23:
Super-seeded by #187

PR #117, opened 2017-12-11
enforce holy import trinity using goimports

databus23:
This PR changes our `gofmt` check to
`goimports -local github.com/sapcc/kubernikus`.

This causes the tests to fail if imports are not ordered in the following way:
```
stdlib
[empty line]
external deps
[empty line]
kubernikus internal deps
````

I purposly excluded the generated files from this check to avoid confusion when we regenerate stuff.

I added y'all as reviewers because I want everyone to be aware of this change.


BugRoger:
resolved merge conflicts

PR #118, opened 2017-12-11
Request-ID Tracing

BugRoger:
This adds a middleware that generates a UUID on each incoming API request. For a poor-man's tracing this `id` is passed as value in a request specific logger.

### DIFF
--- a/pkg/util/log/middleware.go
+++ b/pkg/util/log/middleware.go
@@ -2,6 +2,7 @@ package log
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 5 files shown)


pkg/util/log/middleware.go
```diff
@@ -2,22 +2,41 @@ package log
...
 	"net/http"
 	"strings"
 	"time"
 
 	kitlog "github.com/go-kit/kit/log"
+	uuid "github.com/satori/go.uuid"
 )
 
+const (
+	KubernikusRequestID = "Kubernikus-Request-Id"
```
databus23: Quoting the golang docs for `WithValue`:

> The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. Users of WithValue should define their own types for keys. ...

databus23: The original blog post about `Context` gives an example on how they suggest this shut be used: https://blog.golang.org/context#TOC_3.2.

pkg/util/log/middleware.go
```diff
@@ -2,22 +2,41 @@ package log
...
+)
+
+func RequestIDHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, request *http.Request) {
+		if id := request.Context().Value(KubernikusRequestID); id == nil {
+			request = request.WithContext(context.WithValue(request.Context(), KubernikusRequestID, uuid.NewV4()))
```
databus23: For my taste there we should have the world `context` in this line more often. ;)

pkg/util/log/middleware.go
```diff
@@ -2,22 +2,41 @@ package log
...
+
 func LoggingHandler(logger kitlog.Logger, next http.Handler) http.Handler {
-	logger = kitlog.With(logger, "api", "ingress")
+	ingress_logger := kitlog.With(logger, "api", "ingress")
 	return http.HandlerFunc(func(rw http.ResponseWriter, request *http.Request) {
 		wrapper := makeWrapper(rw)
 
+		id := fmt.Sprintf("%s", request.Context().Value(KubernikusRequestID))
+		request = request.WithContext(context.WithValue(request.Context(), "logger", kitlog.With(logger, "id", id)))
+
```
databus23: I know it annoying to handle this but we should handle the case were the there is no kubernikus request id in the content (because the requestid middleware is disabled).

pkg/util/log/middleware.go
BugRoger: Thanks, good tip. Much better...

pkg/util/log/middleware.go
BugRoger: I'm omitting the `id` key now if the context does not contain the requestId.

pkg/util/log/middleware.go
BugRoger: i resisted the temptation to make this wish come true ;)

BugRoger:
rebased and squashed

PR #125, opened 2017-12-15
removes glog 

BugRoger:
This commit removes all glog references in favor of the structured
logging approach using go-kit/log.

There is plenty room for improvement. The intention here is to rid glog
without introducing to many bugs.

`kubernikusctl` is left to use glog as the main intent of logging here
is human consumption on the cli without an intermediate log retrieval
system. Also, there's tight integration with k8s clients that log using
glog anyway.

The log output is not yet completely clean and 100% structured logging,
because of the informer framwork being uncooperative. We could get rid
of that using a higher default log threshold instead of the default
`INFO`. That exercise is left for another time.

### DIFF
--- a/pkg/api/handlers/create_cluster.go
+++ b/pkg/api/handlers/create_cluster.go
@@ -3,7 +3,6 @@ package handlers
 import (
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/go-openapi/validate"
-	"github.com/golang/glog"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -24,6 +23,7 @@ type createCluster struct {
 }
 
 func (d *createCluster) Handle(params operations.CreateClusterParams, principal *models.Principal) middleware.Responder {
+	logger := getTracingLogger(params.HTTPRequest)
 	name := params.Body.Name
 	spec := params.Body.Spec
 
@@ -49,7 +49,12 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal 
 	k8sutil.EnsureNamespace(d.Kubernetes, d.Namespace)
 	kluster, err = d.Kubernikus.Kubernikus().Klusters(d.Namespace).Create(kluster)
 	if err != nil {
-		glog.Errorf("Failed to create cluster: %s", err)
+		logger.Log(
+			"msg", "failed to create cluster",
+			"kluster", kluster.GetName(),
+			"project", kluster.Account(),
+			"err", err)
+
 		if apierrors.IsAlreadyExists(err) {
 			return NewErrorResponse(&operations.CreateClusterDefault{}, 409, "Cluster with name %s already exists", name)
 		}
--- a/pkg/client/helm/client.go
+++ b/pkg/client/helm/client.go
@@ -5,7 +5,7 @@ import (
 	"os"
 	"runtime"
 
-	"github.com/golang/glog"
+	"github.com/go-kit/kit/log"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/helm/pkg/helm"
@@ -13,7 +13,8 @@ import (
 	"github.com/sapcc/kubernikus/pkg/client/helm/portforwarder"
 )
 
-func NewClient(kubeClient kubernetes.Interface, kubeConfig *rest.Config) (*helm.Client, error) {
+func NewClient(kubeClient kubernetes.Interface, kubeConfig *rest.Config, logger log.Logger) (*helm.Client, error) {
+	logger = log.With(logger, "client", "helm")
 
 	tillerHost := os.Getenv("TILLER_DEPLOY_SERVICE_HOST")
 	if tillerHost == "" {
@@ -26,7 +27,9 @@ func NewClient(kubeClient kubernetes.Interface, kubeConfig *rest.Config) (*helm.
 	tillerHost = fmt.Sprintf("%s:%s", tillerHost, tillerPort)
 
 	if _, err := rest.InClusterConfig(); err != nil {
-		glog.V(2).Info("We are not running inside the cluster. Creating tunnel to tiller pod.")
+		logger.Log(
+			"msg", "We are not running inside the cluster. Creating tunnel to tiller pod.",
+			"v", 2)
 		tunnel, err := portforwarder.New("kube-system", kubeClient, kubeConfig)
 		if err != nil {
 			return nil, err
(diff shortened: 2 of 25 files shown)


pkg/api/handlers/create_cluster.go
```diff
@@ -49,7 +49,8 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal
 	k8sutil.EnsureNamespace(d.Kubernetes, d.Namespace)
 	kluster, err = d.Kubernikus.Kubernikus().Klusters(d.Namespace).Create(kluster)
 	if err != nil {
-		glog.Errorf("Failed to create cluster: %s", err)
+		logger.Log("msg", "failed to create cluster", "err", err)
```
databus23: Not sure if this already part of the context but maybe we should add the cluster name here?

pkg/client/helm/client.go
```diff
@@ -26,7 +27,7 @@ func NewClient(kubeClient kubernetes.Interface, kubeConfig *rest.Config) (*helm.
 	tillerHost = fmt.Sprintf("%s:%s", tillerHost, tillerPort)
 
 	if _, err := rest.InClusterConfig(); err != nil {
-		glog.V(2).Info("We are not running inside the cluster. Creating tunnel to tiller pod.")
+		logger.Log("msg", "We are not running inside the cluster. Creating tunnel to tiller pod.")
```
databus23: level omitted intentionally?

pkg/client/helm/client.go
```diff
@@ -35,7 +36,7 @@ func NewClient(kubeClient kubernetes.Interface, kubeConfig *rest.Config) (*helm.
 		client := helm.NewClient(helm.Host(tillerHost))
 		//Lets see how this goes: We close the tunnel as soon as the client is GC'ed.
 		runtime.SetFinalizer(client, func(_ *helm.Client) {
-			glog.V(2).Info("Tearing Down tunnel to tiller at %s", tillerHost)
+			logger.Log("msg", "trearing down tunnel to tiller", "host", tillerHost)
```
databus23: level omitted intentionally?

pkg/client/openstack/client.go
```diff
@@ -523,7 +535,10 @@ func (c *client) GetNodes(kluster *kubernikus_v1.Kluster, pool *models.NodePool)
 	err = servers.List(client, opts).EachPage(func(page pagination.Page) (bool, error) {
 		nodes, err = ExtractServers(page)
 		if err != nil {
-			glog.V(5).Infof("Couldn't extract server %v", err)
+			c.Logger.Log(
```
databus23: Maybe we should log this errors here as the function returns this error to the caller.

pkg/client/openstack/client.go
```diff
@@ -536,7 +551,22 @@ func (c *client) GetNodes(kluster *kubernikus_v1.Kluster, pool *models.NodePool)
...
+	var name string
+
+	defer func() {
+		c.Logger.Log(
+			"msg", "created node",
+			"kluster", kluster.Name,
+			"project", kluster.Account(),
+			"name", name,
+			"id", id,
+			"err", err,
```
databus23: the `TrailingNilFilter` will not remove this because of the following `v` key. If we add an `err` field that might be nil, it should be the last key.

pkg/client/openstack/client.go
```diff
@@ -560,13 +587,23 @@ func (c *client) CreateNode(kluster *kubernikus_v1.Kluster, pool *models.NodePoo
...
 
-func (c *client) DeleteNode(kluster *kubernikus_v1.Kluster, ID string) error {
+func (c *client) DeleteNode(kluster *kubernikus_v1.Kluster, ID string) (err error) {
+	defer func() {
+		c.Logger.Log(
+			"msg", "deleted node",
+			"kluster", kluster.Name,
+			"project", kluster.Account(),
+			"id", ID,
+			"err", err,
```
databus23: Same as above, better put this last.

pkg/controller/operator.go
```diff
@@ -144,42 +143,48 @@ func NewKubernikusOperator(options *KubernikusOperatorOptions, logger log.Logger
...
-	o.Clients.Satellites = kube.NewSharedClientFactory(secrets, klusters)
+	o.Clients.Satellites = kube.NewSharedClientFactory(secrets, klusters, logger)
 
 	// Add kubernikus types to the default Kubernetes Scheme so events can be
 	// logged for those types.
 	v1.AddToScheme(scheme.Scheme)
 	eventBroadcaster := record.NewBroadcaster()
-	eventBroadcaster.StartLogging(glog.Infof)
+	eventBroadcaster.StartLogging(func(format string, args ...interface{}) {
+		o.Logger.Log("msg", fmt.Sprint(format, args), "v", 2)
```
databus23: We could potentially log events more structured as the arguments always mean the same. We should at least some key that allows us to filter for events.
I'm also fine with leaving this for a follow up.

pkg/templates/ignition.go
```diff
@@ -90,37 +90,43 @@ func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, secret *v1.Secret
...
+	var report report.Report
+
+	defer func() {
+		logger.Log(
+			"msg", "ignition debug",
+			"data", data,
+			"yaml", string(buffer.Bytes()),
+			"json", string(dataOut),
+			"report", report.String(),
+			"err", err,
```
databus23: Same as above, better put this last.

pkg/api/handlers/create_cluster.go
BugRoger: Added `kluster` and `project` values.

pkg/client/helm/client.go
BugRoger: Overlooked that level. Added it again.

pkg/client/helm/client.go
BugRoger: Same as above.

pkg/client/openstack/client.go
BugRoger: Yeah, remove that.

pkg/client/openstack/client.go
BugRoger: Ah, I didn't know it works like that. I thought it just checks for `err ==  nil`

pkg/client/openstack/client.go
BugRoger: switched

pkg/templates/ignition.go
BugRoger: also switched

pkg/controller/operator.go
BugRoger: Much better.

```
ts=2017-12-18T15:06:19.787737058Z caller=github.com/sapcc/kubernikus/pkg/controller/operator.go:156 
controller=operator 
resource=event 
msg="Successfully created node 2657e86b-be51-4e5c-a5a5-dce9c5431661"
reason=SuccessfullCreateNode 
type=Normal 
kind=Kluster 
namespace=kubernikus-michi 
name=test2-5d725ddf97664a16b011e8a8dd75772b 
v=2
```

BugRoger: Much better.

```
ts=2017-12-18T15:06:19.787737058Z caller=github.com/sapcc/kubernikus/pkg/controller/operator.go:156 
controller=operator 
resource=event 
msg="Successfully created node 2657e86b-be51-4e5c-a5a5-dce9c5431661"
reason=SuccessfullCreateNode 
type=Normal 
kind=Kluster 
namespace=kubernikus-michi 
name=test2-5d725ddf97664a16b011e8a8dd75772b 
v=2
```

BugRoger:
Squashed and Rebased. Ready for merge as soon as Travis and @SchwarzM approve!

PR #126, opened 2017-12-15
Specify security groups by id when creating servers

databus23:
This PR adds a field `SecurityGroupID` to the kluster spec.
If its not specified the auto discovery fills in the `id` of the `default` security group.
This id is then used when creating instances (e.g. by launchctl).

I added a poor mans migration into the `groundctl` reconciliation to retrofit existing clusters with the `id` of the projects `default` security group.
I will remove this hack in a subsequent commit once it has been deployed everywhere (and all clusters have the security group set)

### DIFF
--- a/pkg/controller/ground.go
+++ b/pkg/controller/ground.go
@@ -141,6 +141,11 @@ func (op *GroundControl) handler(key string) error {
 		metrics.SetMetricKlusterInfo(kluster.GetNamespace(), kluster.GetName(), kluster.Status.Version, kluster.Spec.Openstack.ProjectID, kluster.GetAnnotations(), kluster.GetLabels())
 		metrics.SetMetricKlusterStatusPhase(kluster.GetName(), kluster.Status.Phase)
 
+		//TODO: remove ASAP, this is just a poor mans migration, adding the sec group id to existing klusters
+		if err := op.ensureSecurityGroupID(kluster); err != nil {
+			op.Recorder.Eventf(kluster, api_v1.EventTypeWarning, ConfigurationError, "Failed to add default security grop id to kluster: %s", err)
+		}
+
 		switch phase := kluster.Status.Phase; phase {
 		case models.KlusterPhasePending:
 			{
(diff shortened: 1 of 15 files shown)


pkg/controller/ground.go
```diff
@@ -141,6 +141,11 @@ func (op *GroundControl) handler(key string) error {
 		metrics.SetMetricKlusterInfo(kluster.GetNamespace(), kluster.GetName(), kluster.Status.Version, kluster.Spec.Openstack.ProjectID, kluster.GetAnnotations(), kluster.GetLabels())
 		metrics.SetMetricKlusterStatusPhase(kluster.GetName(), kluster.Status.Phase)
 
+		//TODO: remove ASAP, this is just a poor mans migration, adding the sec group id to existing klusters
```
BugRoger: famous last words...

pkg/controller/ground.go
```diff
@@ -477,10 +482,40 @@ func (op *GroundControl) discoverOpenstackInfo(kluster *v1.Kluster) error {
 		}
 	}
 
+	if securityGroupID := copy.Spec.Openstack.SecurityGroupID; securityGroupID != "" {
+		//TODO: Validate that the securitygroup id exists
```
BugRoger: forgot to add this?

pkg/controller/ground.go
```diff
@@ -477,10 +482,40 @@ func (op *GroundControl) discoverOpenstackInfo(kluster *v1.Kluster) error {
...
+
+	} else {
+		id, err := op.Clients.Openstack.GetSecurityGroupID(kluster.Account(), "default")
+		if err != nil {
+			return fmt.Errorf("Failed to get id for default securitygroup in project %s: %s", err, kluster.Account())
+		}
+		glog.V(5).Infof("[%v] Setting SecurityGroupID to %v", kluster.Name, copy.Spec.Openstack.SecurityGroupID)
```
BugRoger: Nooooo.... 😄

pkg/controller/ground.go
databus23: I forgot this intentionally. For this I would have needed to add yet another method to the old openstack client (`GetSecurityGroup(id string)`) which I refused to do. But maybe its better to add it now, so that when we (you) refactor the client all required method are already present.

Issue #114, opened 2017-12-09
Monitor wormhole tunnel status


databus23:
I just had to restart a wormhole client on a node for the first time to fix a node.
That got me thinking we need to monitor the tunnel status for each node because the error behaviour with a broken tunnel is mysterious and not immediately visible:
* pods can't reach the kubernetes api anymore (`EOF`)
* `kubectl` `exec`, `logs`,`proxy` does not work anymore

My current thinking is we should add a custom `NodeCondition`, maybe `TunnelUnavilable`?
We can then bubble this up via our node watcher @auhlig is working on.

**UPDATE:**
The wormhole server which runs as sidecar to the apiserver should monitor the connection state for each node and update the `NodeCondition` accordingly.

Tasks:
- [ ] Modify wormhole server library to expose connection states
- [ ] Update the wormhole controller to reconcile the `TunnelUnavailable` Condition for each node be looking up the connection status in the wormhole server for each node in the informer
- [ ] Figure out how to upgrade the wormhole server image in running Klusters (as we don't have a general upgrade feature yet). Maybe using #182

databus23:
Digging through node problem detector, this seems to be the way to update conditions efficiently:
https://github.com/kubernetes/node-problem-detector/blob/ec470b65b045eee65d55df4ed243a801ce5953d8/pkg/problemclient/problem_client.go#L89

auhlig:
+1 for the node-problem-detector. I did this for the kernel deadlock error once. Isn't too hard if we have a scriptable way to monitor the tunnel status

databus23:
@auhlig I'm sorry I somehow missed that you took my mentioning of node-problem-detector that we should use it for monitoring the tunnel status. That is not what I meant. I just wanted to point out how to update a `NodeCondition` efficiently using `client-go`.
Here is my reasoning why I don't want to use npd:

1. As already mentioned in #127 using npd as a pod in the target cluster is problematic because then npd relies on the tunnel for apiserver access, which means it can't update the condition when the tunnel is unavailable.
2. I'm of the opinion that we should not "litter" the target cluster with infrastructure pods to help us do our work. The cluster belongs to the user and it should be as pristine as possible. Currently we only scheduled kube-dns in the cluster because we can't easily pull it out. We should try to not add more to the target cluster. Running node-problem-dector in the cluster should be a decision left to the user. If he wants to have it, he can install it potentially with his own custom checks for whatever.

[...]

PR #127, opened 2017-12-15
**WIP** monitor wormhole tunnel

auhlig:
As mentioned in #114, use node problem detector (npd) to monitor tunnel status and maintain as condition `KubernikusTunnelStatus` on each node. This is the simple approach relying on a scriptable way to check the status, if feasible. See comment below.
NPD runs as a ds in each kluster and will be part of kube-master chart.

### DIFF
--- a/charts/kube-master/charts/node-problem-detector/check_tunnel.sh
+++ b/charts/kube-master/charts/node-problem-detector/check_tunnel.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+OK=0
+NONOK=1
+UNKNOWN=2
+
+# TODO: how to check if tunnel is up
+
+# TODO: cmd to verify tunnel status
+#if [ $? -ne 0 ]; then
+#    echo "Systemd is not supported"
+#    exit $UNKNOWN
+#fi
+
+# TODO: cmd to verify tunnel status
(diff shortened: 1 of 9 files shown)


charts/kube-master/charts/node-problem-detector/check_tunnel.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+OK=0
+NONOK=1
+UNKNOWN=2
+
+# TODO: how to check if tunnel is up
```
auhlig: just the rough structure for the script. need to discuss how to actually monitor tunnel status

charts/kube-master/charts/node-problem-detector/kubernikus-tunnel.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "source": "ntp-custom-plugin-monitor",
+  "conditions": [
+    {
+      "type": "KubernikusTunnelStatus",
+      "reason": "TunnelIsUP",
+      "message": "tunnel is up"
+    }
+  ],
+  "rules": [
```
auhlig: permanent or temporary issue? i think the latter, right?

auhlig:
Ah. I see.

PR #129, opened 2017-12-19
Add leveled logging

databus23:
I reused the `—v` flag for glog to set the desired log level. By default we only log messages with no level or level 0.
For the flag processing to work I had to move the logger creation into the cobra commands `Run` functions because otherwise I can’t access the flag without hussle.

I moved the logging setup to logutil.NewLogger() to that we can reuse that function in all binaries.

### DIFF
--- a/cmd/apiserver/main.go
+++ b/cmd/apiserver/main.go
@@ -24,16 +24,13 @@ func init() {
 }
 
 func main() {
-	var logger kitLog.Logger
-	logger = kitLog.NewLogfmtLogger(kitLog.NewSyncWriter(os.Stderr))
-	logger = logutil.NewTrailingNilFilter(logger)
-	logger = kitLog.With(logger, "ts", kitLog.DefaultTimestampUTC, "caller", Caller(3))
+	if f := goflag.Lookup("logtostderr"); f != nil {
+		f.Value.Set("true") // log to stderr by default
+	}
 
 	swaggerSpec, err := spec.Spec()
 	if err != nil {
-		logger.Log(
-			"msg", "failed to spec swagger spec",
-			"err", err)
+		fmt.Printf(`failed to parse swagger spec: %s`, err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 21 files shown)


cmd/apiserver/main.go
```diff
@@ -52,12 +49,10 @@ func main() {
 		fmt.Fprintln(os.Stderr, pflag.CommandLine.FlagUsages())
 	}
 	// parse the CLI flags
-	if f := goflag.Lookup("logtostderr"); f != nil {
-		f.Value.Set("true") // log to stderr by default
-	}
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine) //slurp in glog flags
 	pflag.Parse()
 	//goflag.CommandLine.Parse([]string{}) //https://github.com/kubernetes/kubernetes/issues/17162
+	logger := logutil.NewLogger(pflag.CommandLine)
```
BugRoger: The `Caller` func below can be deleted with this refactoring.

pkg/cmd/kubernikus/operator.go
```diff
@@ -91,10 +86,8 @@ func (o *Options) Complete(args []string) error {
 }
 
 func (o *Options) Run(c *cobra.Command) error {
-	var logger log.Logger
-	logger = log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
-	logger = logutil.NewTrailingNilFilter(logger)
-	logger = log.With(logger, "ts", log.DefaultTimestampUTC, "caller", Caller(3))
+
+	logger := logutil.NewLogger(c.Flags())
```
BugRoger: The Caller func below can be deleted with this refactoring.

pkg/wormhole/server.go
```diff
@@ -37,9 +35,6 @@ type Server struct {
 
 func NewServer(options *ServerOptions) (*Server, error) {
 	s := &Server{Logger: log.With(options.Logger, "wormhole", "server")}
-	if options.TunnelOptions.Logger == nil {
```
BugRoger: Isn't the logger in TunnelOptions now `nil` again?

cmd/apiserver/main.go
databus23: Good catch

pkg/cmd/kubernikus/operator.go
databus23: Yes, yes

pkg/wormhole/server.go
databus23: No because I removed the logger from the our ServerOptions struct.
But while looking at the code I got dissatisfied with it. I think it was kind of convoluted due to fact that we split the work on it you writing the controller and my writing the tunnel server.
I'm taking the opportunity to refactor the wormhole server a bit to de-obfiuscate the options handling a little bit.

PR #132, opened 2017-12-19
Upgrade to Kubernetes 1.8.5

### LINKED ISSUE #62: Upgrade to 1.8
Planed release for 1.8 is Sept 27th (Weds): v1.8.0 release!

Uhm, tomorrow. It's still in v1.8.0-rc1 but we can start upgrading already. Let's upgrade before we go Beta.

BugRoger:
This commit upgrades the helm chart and ignition template to use 1.8.5.

Currently, this assumes that klusters without version information in the
spec are on 1.7.5. All new klusters must provide a version.

With 1.8.5 the default behaviour of the Openstack LoadBalancer creation
changed. Instead of defaulting to an internal LoadBalancer it now
creates an external LB by default. This requires the external network to
be configured in the cloud-provider config.

Fixes #62


BugRoger:
Before anyone asks - the Gurkenwurst hack is still required for foreseeable time.

https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L212

That anyone runs this outside of a VM seems to blow a few assumptions.

pkg/templates/node_v1.8.go
```diff
@@ -0,0 +1,378 @@
+/* vim: set filetype=yaml : */
```
SchwarzM: The node 1.7 and 1.8 template look very similar is it currently possible to extract the common stuff and merge in the kubernetes version? 
Changes to the common coreos setup would apply to all kubernetes versions.

BugRoger: Yeah, I thought about that too. I'd do it in a separate PR.

charts/kube-master/templates/_openstack.config.tpl
```diff
@@ -13,6 +13,7 @@ region = {{ .Values.openstack.region }}
 [LoadBalancer]
 lb-version=v2
 subnet-id= {{ required "missing openstack.lbSubnetID" .Values.openstack.lbSubnetID }}
+floating-network-id= {{ required "missing openstack.lbFloatingNetworkID" .Values.openstack.lbFloatingNetworkID }}
```
databus23: for 1.7 this is not strictly *required*. But maybe we collect it any case.

charts/kube-master/templates/api.yaml
```diff
@@ -107,9 +107,14 @@ spec:
             - --authorization-mode=Node,RBAC
             - --cloud-config=/etc/kubernetes/cloudprovider/openstack.config
             - --cloud-provider=openstack
+{{- if hasPrefix .Values.version.kubernetes "v1.8" }}
```
databus23: I don't see a point in having the *v* prefix in our values for the kubernetes version and would find it clear to just have the version number for comparison.

charts/kube-master/values.yaml
```diff
@@ -29,7 +30,10 @@ clusterCIDR: 198.19.0.0/16
 serviceCIDR: 198.18.128.0/17
 advertiseAddress: 198.18.128.1
 #bootstrapToken
-#version:
+
+version:
+# kubernikus:
+  kubernetes: v1.8.5
```
databus23: As said above I would go without the *v* for this field

BugRoger:
@databus23 My thought about the versions were to just use a git ref. For Kubernikus we use the Git commit and in Kubernetes the releases are tagged including `v`

image

image

Also the images include the `v` too:

image

Though I don't have a super strong opinion either way.

databus23:
OK, I get that. I still like it moreif the version field in the `Spec` (or `Status`) would just contain the version without prefix or suffix but its not that important to argue about. I'm fine with it either way.

charts/kube-master/templates/_openstack.config.tpl
BugRoger: I didn't want to litter too many `if`s throughout the code. De facto `1.7` won't be used anymore anyway...

BugRoger:
Removed the `v` and added poor-man's CoreOS versioning scheme generator

Issue #133, opened 2017-12-20
Get cluster that doesn't exist gives 502 instead of 404


edda:
Trying to get a cluster that doesn't exist anymore (`/api/v1/clusters/bla`) results in a 502 Bad Gateway response now. This should (and used to) return 404 instead.

databus23:
Hmm can't reproduce locally. I get the expected:
```
curl -i http://localhost:5100/api/v1/clusters/bla
HTTP/1.1 404 Not Found
Content-Type: application/json
Vary: Origin
Date: Wed, 20 Dec 2017 13:36:01 GMT
Content-Length: 35

{"code":404,"message":"Not found"}
```

databus23:
ok I can reproduce in eu-nl-1

databus23:
It seems the nginx-ingress is trying to redirect 404 to the default backend (which it shouldn't) and that fails with
```
2017/12/20 13:41:55 [error] 3302#3302: *116611 upstream prematurely closed connection while reading response header from upstream, client: 198.18.0.76, server: [REDACTED], request: "GET /api/clusters HTTP/2.0", upstream: "http://192.168.0.73:8080/", host: " [REDACTED]"
```
That should fail either (192.168.0.73 is the default backend)

databus23:
the default backend is failing with:
```
2017/12/20 13:41:55 http: panic serving 192.168.0.72:36388: runtime error: index out of range
goroutine 104487 [running]:
net/http.(*conn).serve.func1(0xc42008a8c0)
	/usr/local/opt/go/libexec/src/net/http/server.go:1697 +0xd0
panic(0x6d4520, 0x8d3ae0)
	/usr/local/opt/go/libexec/src/runtime/panic.go:491 +0x283
main.errorHandler.func1(0x8a9660, 0xc4201182a0, 0xc420166300)
	/Users/d064935/.go/src/k8s.io/ingress-nginx/images/custom-error-pages/main.go:69 +0xeba
net/http.HandlerFunc.ServeHTTP(0xc42011a0e0, 0x8a9660, 0xc4201182a0, 0xc420166300)
	/usr/local/opt/go/libexec/src/net/http/server.go:1918 +0x44
net/http.(*ServeMux).ServeHTTP(0x8de700, 0x8a9660, 0xc4201182a0, 0xc420166300)
	/usr/local/opt/go/libexec/src/net/http/server.go:2254 +0x130
[...]

PR #134, opened 2017-12-20
fix default backend and use it only for 502,503,504

auhlig:
The `out of range error` happend [here](https://github.com/kubernetes/ingress-nginx/blob/master/images/custom-error-pages/main.go#L64). Added a check before.
Also just use the default backend for `502,503,504`.
Issue https://github.com/sapcc/kubernikus/issues/133 /cc @databus23


databus23:
What do you mean by

> Added a check before.

You linked a kubernetes repo. Did you fork it? Where? Should we maybe file this upstream.

auhlig:
Yep, I forked it. Didn't push it so far. I'll do in a min.

databus23:
@auhlig Would you care to try getting this merged upstream?

auhlig:
will do

PR #136, opened 2017-12-20
Preeseed rolebinding for kubernikus:member

jonathanschwarze:
Add RoleBinding in Namespace "default" for kubernikus:member


pkg/controller/ground/bootstrap.go
```diff
@@ -140,6 +163,19 @@ func CreateOrUpdateClusterRoleBinding(client clientset.Interface, clusterRoleBin
 	return nil
 }
 
+func CreateOrUpdateRoleBinding(client clientset.Interface, RoleBinding *rbac.RoleBinding) error {
```
BugRoger: lowercase `RoleBinding` plx

pkg/controller/ground/bootstrap.go
```diff
@@ -140,6 +163,19 @@ func CreateOrUpdateClusterRoleBinding(client clientset.Interface, clusterRoleBin
...
+	if _, err := client.RbacV1beta1().RoleBindings("default").Create(RoleBinding); err != nil {
```
BugRoger: Isn't it possible to infer the `default` Namespace from the `rolebinding.Meta`?

pkg/controller/ground/bootstrap.go
```diff
@@ -140,6 +163,19 @@ func CreateOrUpdateClusterRoleBinding(client clientset.Interface, clusterRoleBin
...
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("unable to create RBAC rolebinding: %v", err)
+		}
+
+		if _, err := client.RbacV1beta1().RoleBindings("default").Update(RoleBinding); err != nil {
```
BugRoger: same as above

BugRoger:
Thanks, looks good now. I think, you should checkout master and cherry-pick the two commits, to get rid of all the docu that bled in.

BugRoger:
Travis is also a bit unhappy...

PR #137, opened 2017-12-21
Sort (and filter) flavors in metadata response

### LINKED ISSUE #130: Metadata: Sort and filter flavors
For the flavors in the metadata API call I suggest the following changes:

1) Sort flavors by RAM first, VCPUs second
2) Delete flavors from the list which we don't support (because they are too small or for other reasons).

databus23:
Flavors are now sorted by RAM and VCPUs.
Also the RAM and VCPUs are now part of the reponse.
Additionally I hardcoded the flavors list call for now to only return flavors with 2000 MB RAM or more to get rid of the tiny images that don’t work with our default coreos image

Fixes #130

### DIFF
--- a/pkg/api/models/flavor_sorter.go
+++ b/pkg/api/models/flavor_sorter.go
@@ -1,0 +1,29 @@
+package models
+
+import (
+	"sort"
+)
+
+type flavorSorter struct {
+	flavors []Flavor
+}
+
+func SortFlavors(flavors []Flavor) {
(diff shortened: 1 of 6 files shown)


BugRoger:
@edda When this is merged the `Flavors` model changes to include two additional attributes? Does this need a coordinated rollout with the UI?

https://github.com/sapcc/kubernikus/pull/137/files#diff-65058f891490afbb80c8704444e0af9bR216

BugRoger:
Merging away then 👍

PR #144, opened 2018-01-02
Adds Kluster Deorbit Controller

### LINKED ISSUE #45: Kluster Garbage Collector
When a kluster is deleted we need a controlled demolition that tears down all Openstack objects that have been created.

- Routes
- LoadBalancers, Listeners, Pools
- Volumes

BugRoger:
This adds a controller that does a sequenced decomissioning of Klusters. Fixes #45.

For an initial version the most pressing issue is that the Kluster's ControllerManager is just being terminated. This doesn't give the Kubernetes controllers enough time to remove remaining OpenStack resources. Therefore, trash just keeps on collecting in the user's project. 

To make matters worse, manual clean-up is not straight forward:

* Load Balancers require a whole deletion sequence
* Volumes take a non-trivial amount of time to actually detach/delete
* Identifying which LBs/Volumes belong to the kluster or other resources in the project is not always possible
* If multi-kluster support per project becomes a reality, it will be next to impossible for the user to distinguish which resource belongs to which cluster.

Given that, my approach in this PR:

1. Delete all Services of `type=LoadBalancer`
2. Delete all PVCs
3. Wait until everything is deleted

It turns out step 3 is tricky. Deleting services happens asynchronously. The actual observable resources are deleted immediately. While the controller-manager in the background keeps cached retry-loops active. This is not observable from the outside. See: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/service/service_controller.go#L801

We could check the resources in Openstack and continue once they are gone. It's not easy to figure out what belongs to a kluster and to the project though. 

Currently, i'm approaching this by adding a wait period. For Klusters without LBs and Volumes, 1min seems sufficient for the RouteController. For LBs we wait for 2minutes. Deletion might happen sequentially and that needs to be increased dynamically.

### DIFF
--- a/docs/development/controllers.md
+++ b/docs/development/controllers.md
@@ -2,6 +2,37 @@ Kubernikus Controllers
 ======================
 The Kubernikus operator contains various independent controllers each one handling a different aspect of the managed kubernetes clusters.
 
+Deorbit Controller
+------------------
+
+The Kubernetes `controller-manager` runs controllers that are responsible for
+creating, updating and deleting of OpenStack resources. It auto-provisions
+Cinder volumes for Persistent Volume Claims, load balancers for services and
+routes for nodes in the cluster.  When a kluster is terminated it is desired
+that these resources are cleaned up automatically. 
+
+The problem is that the `controller-manager` only cleans up, when the
+respective resources are deleted from the cluster.
+
+This is where the `deorbiter` comes into play. It inspects the cluster and
+deletes all PVCs that are backed by Cinder, as well as all services of type
+`LoadBalancer`. Afterwards, it waits until the `controller-manager` has
+completed the clean-up.
+
+Problematic is that Cinder volumes can take a variable time until they are
+deleted. In some cases volumes might be stuck deleting indefinitely. The
+`deorbiter` watches the PVs (note: PV not PVC) in the cluster and waits for all
+of them to disappear. PVs are guarded by a finalizer and are only removed when
+the cleanup completes. The `deorbiter` retries the deletion for a configurable
+amount of time and eventually self-destructs the cluster. This might leave some
+debris in the OpenStack project. 
+
+For services Kubernetes does currently not implement the same finalizer
+concept. That means that a service is deleted immediately while the clean-up
+progresses in the background. Without double checking directly in OpenStack,
+the `deboriter` can't tell when the deletion is completed. Here we assume
+a fixed time, like 2 minutes, and hope for the best. 
+
 
 Route garbage collector
 -----------------------
@@ -11,3 +42,5 @@ The bug occasionally leaves static route entries in the cluster's OpenStack rout
 The bug is fixed in the upcoming 1.10 version of kubernetes. That means this controller is only needed for clusters <1.10.
 
 For each cluster the controller polls the corresponding OpenStack router and inspects the configured static routes. It tries to identify and remove orphaned routes to fix the clusters networking. It removes all route entries for CIDR ranges that are within the clusters CIDR range for Pods where the target/nextHop ip address can't be matched to an OpenStack compute instance.
+
+
--- a/pkg/api/handlers/terminate_cluster.go
+++ b/pkg/api/handlers/terminate_cluster.go
@@ -23,19 +23,12 @@ type terminateCluster struct {
 func (d *terminateCluster) Handle(params operations.TerminateClusterParams, principal *models.Principal) middleware.Responder {
 
 	kluster := d.Kubernikus.Kubernikus().Klusters(d.Namespace)
-	k, err := kluster.Get(qualifiedName(params.Name, principal.Account), metav1.GetOptions{})
+	_, err := kluster.Get(qualifiedName(params.Name, principal.Account), metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			return NewErrorResponse(&operations.TerminateClusterDefault{}, 404, "Not found")
 		}
 		return NewErrorResponse(&operations.TerminateClusterDefault{}, 500, err.Error())
-	}
-	if k.Status.NodePools != nil && len(k.Status.NodePools) > 0 {
-		for _, nodepoolinfo := range k.Status.NodePools {
-			if nodepoolinfo.Running > 0 {
-				return NewErrorResponse(&operations.TerminateClusterDefault{}, 409, "Cluster still has Nodes in a Pool")
-			}
-		}
 	}
 
 	_, err = editCluster(kluster, principal, params.Name, func(kluster *v1.Kluster) {
--- a/pkg/controller/deorbit/controller.go
+++ b/pkg/controller/deorbit/controller.go
@@ -1,0 +1,149 @@
+package deorbit
+
+import (
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/sapcc/kubernikus/pkg/api/models"
+	"github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
+	"github.com/sapcc/kubernikus/pkg/controller/base"
+	"github.com/sapcc/kubernikus/pkg/controller/config"
+	"github.com/sapcc/kubernikus/pkg/controller/metrics"
+	informers_kubernikus "github.com/sapcc/kubernikus/pkg/generated/informers/externalversions/kubernikus/v1"
+	"github.com/sapcc/kubernikus/pkg/util"
+)
+
+func init() {
+	prometheus.MustRegister(
+		metrics.DeorbitOperationsLatency,
+		metrics.DeorbitOperationsTotal,
+		metrics.DeorbitSuccessfulOperationsTotal,
+		metrics.DeorbitFailedOperationsTotal,
+	)
+}
+
+const (
+	DeorbiterFinalizer = "deorbiter"
+
+	// The longest time a single worker will be blocked, while waiting for the clean
+	// up of OpenStack resources. After this timeout, deorbiting will be queued again.
+	// It's a safeguard against having all workers hanging indefinetly and congesting
+	// the queue.
+	UnblockWorkerTimeout = 10 * time.Minute
+)
+
+type DeorbitReconciler struct {
+	config.Clients
+
+	Recorder record.EventRecorder
+	Logger   log.Logger
+
+	klusterInformer informers_kubernikus.KlusterInformer
+}
+
+func NewController(threadiness int, factories config.Factories, clients config.Clients, recorder record.EventRecorder, logger log.Logger) base.Controller {
+	logger = log.With(logger, "controller", "deorbit")
+
+	var reconciler base.Reconciler
+	reconciler = &DeorbitReconciler{clients, recorder, logger, factories.Kubernikus.Kubernikus().V1().Klusters()}
+	reconciler = &base.LoggingReconciler{reconciler, logger}
+	reconciler = &base.InstrumentingReconciler{
+		reconciler,
+		metrics.DeorbitOperationsLatency,
+		metrics.DeorbitOperationsTotal,
+		metrics.DeorbitSuccessfulOperationsTotal,
+		metrics.DeorbitFailedOperationsTotal,
+	}
+	return base.NewController(threadiness, factories, clients, reconciler, logger)
+}
+
+func (d *DeorbitReconciler) Reconcile(kluster *v1.Kluster) (bool, error) {
+	switch kluster.Status.Phase {
+	case models.KlusterPhaseRunning:
+		return false, util.EnsureFinalizerCreated(d.Kubernikus.KubernikusV1(), d.klusterInformer.Lister(), kluster, DeorbiterFinalizer)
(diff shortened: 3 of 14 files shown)


databus23:
I think instead of adding another Phase we should finally take a closer look at finalisers: https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#finalizers

BugRoger:
> I think instead of adding another Kluster phase we should finally take a closer look at finalizers: https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#finalizers

Will remove `KlusterPhaseDeorbitting` in favor of Finalizers as implemented in #155.

pkg/controller/deorbit/deleter_dynamic.go
```diff
@@ -0,0 +1,306 @@
+package deorbit
+
```
SchwarzM: huh?

BugRoger:
This is ready for another review. I made an effort to comment and write readable code. It might be a good review also for @SchwarzM, @auhlig, @vonneudeck, @jonathanschwarze. Would appreciate all your feedback.

pkg/api/handlers/terminate_cluster.go
```diff
@@ -23,22 +22,7 @@ type terminateCluster struct {
...
-	if k.Status.NodePools != nil && len(k.Status.NodePools) > 0 {
-		for _, nodepoolinfo := range k.Status.NodePools {
-			if nodepoolinfo.Running > 0 {
-				return NewErrorResponse(&operations.TerminateClusterDefault{}, 409, "Cluster still has Nodes in a Pool")
-			}
-		}
-	}
-
-	_, err = editCluster(kluster, principal, params.Name, func(kluster *v1.Kluster) {
+	_, err := editCluster(kluster, principal, params.Name, func(kluster *v1.Kluster) {
```
BugRoger: This removes the API block, that disallows Kluster termination if there are still running nodes.

docs/development/controllers.md
```diff
@@ -2,6 +2,37 @@ Kubernikus Controllers
...
+`LoadBalancer`. Afterwards, it waits until the `controller-manager` has
+completed the clean-up.
+
+Problematic is that Cinder volumes can take a variable time until they are
+deleted. In some cases volumes might be stuck deleting indefintelty. The
+`deorbiter` watches the PVs (note: PV not PVC) in the cluster and waits for all
+of them to disappear. PVs are guarded by a finalizer and are only removed when
+the cleanup completes. The `deorbiter` retries the deletion for a configurable
+amount of time and eventually self-destructs the cluster. This might leave some
+debris in the OpenStack project.
```
SchwarzM: can we alert on this situation? i think the whole purpose of deorbiting is defeated when there is a chance of this going unnoticed.

pkg/controller/deorbit/controller.go
```diff
@@ -0,0 +1,149 @@
...
+			return err
+		}
+
+		deletedServices, err := deorbiter.DeleteServices()
+		if err != nil {
+			return err
+		}
+
+		if len(deletedPVCs) > 0 {
+			if err := deorbiter.WaitForPersistentVolumeCleanup(); err != nil {
```
SchwarzM: does this work for manila(nfs) and cinder ? i think regarding the multitude of legacy software that will run we will need to provide some way of read-write-many volumes which would be easiest with manila / nfs.

pkg/controller/deorbit/deorbiter.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if done {
+		return nil
+	}
+
+	return wait.PollUntil(PollInterval, d.isServiceCleanupFinished, d.Stop)
+}
+
+func (d *ConcreteDeorbiter) SelfDestruct(reason SelfDestructReason) (err error) {
+	// Self-Destruct ironically does nothing
```
SchwarzM: at least for deorbit-hanging we should try and "sentry" the hell out of this to provide as much context as possible as accessibly as we can.

pkg/controller/deorbit/events.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return
+}
+
+func (d *EventingDeorbiter) SelfDestruct(reason SelfDestructReason) (err error) {
+	err = d.SelfDestruct(reason)
+
+	if err == nil {
+		d.Recorder.Eventf(d.Kluster, core_v1.EventTypeNormal, events.SuccessfulDeorbitSelfDestruct, "Successfully activated self-destruction. There might be left-over volumes and load balancers.")
```
SchwarzM: if there is debris leftover after this finished successfully i would consider this a bug.

docs/development/controllers.md
```diff
@@ -2,6 +2,37 @@ Kubernikus Controllers
...
+The problem is that the `controller-manager` only cleans up, when the
+respective resources are deleted from the cluster.
+
+This is where the `deorbiter` comes into play. It inspects the cluster and
+deletes all PVCs that are backed by Cinder, as well as all services of type
+`LoadBalancer`. Afterwards, it waits until the `controller-manager` has
+completed the clean-up.
+
+Problematic is that Cinder volumes can take a variable time until they are
+deleted. In some cases volumes might be stuck deleting indefintelty. The
```
databus23: typo: indefinitely

docs/development/controllers.md
```diff
@@ -2,6 +2,37 @@ Kubernikus Controllers
...
+`deorbiter` watches the PVs (note: PV not PVC) in the cluster and waits for all
+of them to disappear. PVs are guarded by a finalizer and are only removed when
+the cleanup completes. The `deorbiter` retries the deletion for a configurable
+amount of time and eventually self-destructs the cluster. This might leave some
+debris in the OpenStack project. 
+
+For services Kubernetes does currently not implement the same finalizer
+concept. That means that a service is deleted immediately while the clean-up
+progresses in the background. Without double checking directly in Openstack,
```
databus23: OpenStack 🤣

pkg/api/handlers/terminate_cluster.go
databus23: Looks to me like there was to much deleted.
Does this still correctly return 404 for non-existent klusters? Unfortunately the `TestClusterUpdate` doesn't test this case...

pkg/controller/deorbit/controller.go
```diff
@@ -0,0 +1,149 @@
...
+	done := make(chan struct{})
+	defer close(done)
+
+	timer := time.NewTimer(UnblockWorkerTimeout)
+	defer timer.Stop()
+
+	go func() {
+		<-timer.C
+		d.Logger.Log("msg", "timeout waiting. unblocking the worker routine")
+		close(done)
```
databus23: Together with line 81 we have a guaranteed panic here when the timeout is hit I think  (close of already closed channel).

pkg/controller/deorbit/events.go
```diff
@@ -0,0 +1,85 @@
...
+	} else {
+		d.Recorder.Eventf(d.Kluster, core_v1.EventTypeNormal, events.FailedDeorbitSelfDestruct, "Failed activate self-destruction. There might be left-over volumes and load balancers.", err)
```
databus23: "Failed *to* activate"

pkg/controller/metrics/deorbit.go
```diff
@@ -0,0 +1,36 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var DeorbitOperationsLatency = prometheus.NewSummaryVec(
```
databus23: I think these need to be registered in an `init` function: See #195

docs/development/controllers.md
BugRoger: fixed

docs/development/controllers.md
BugRoger: Yes, totally. There's a counter being incremented for each self-destruction. We can also see, why it was aborted and how it took. I haven't included the alerts yet as I find that hard to dry-code. Wanted to do that in a dedicated PR.

docs/development/controllers.md
BugRoger: 💩

BugRoger: fixed

BugRoger: 😄

pkg/api/handlers/terminate_cluster.go
BugRoger: Oh, idd. Nice catch. Readded.

pkg/controller/deorbit/controller.go
BugRoger: Hm. You are right. I fixed this now but it feels like there's a simpler way to do that.

pkg/controller/deorbit/controller.go
BugRoger: I thought about that too. The thing is that Manila is not auto-provisioned. The user has to create the PV manually putting in the required parameters. It would be odd, if we would decommission manually created volumes.

So, no. It only works for Cinder. Even if we would delete the Manila PV, nothing would be happening. The cloud-provider only works for Cinder volumes. Or does it? I'm not sure now.

pkg/controller/deorbit/deorbiter.go
BugRoger: Yeah, this does exactly that. This method here is decorated with logs and metrics, so even though we don't do anything but continue, we can alert and measure on the timeouts.

pkg/controller/deorbit/events.go
BugRoger: I reworded the message.

pkg/controller/deorbit/events.go
BugRoger: Thx

pkg/controller/metrics/deorbit.go
BugRoger: Yes, thanks.

BugRoger:
Thanks for the comments. That was very helpful. Fixed all remarks.

pkg/controller/deorbit/deorbiter.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if done {
+		return nil
+	}
+
+	return wait.PollUntil(PollInterval, d.isPersistentVolumesCleanupFinished, d.Stop)
+}
+
+func (d *ConcreteDeorbiter) WaitForServiceCleanup() (err error) {
+	done, err := d.isPersistentVolumesCleanupFinished()
```
databus23: Why is `isPersistentVolumesCleanupFinished` called in `WaitForServiceCleanup`?

pkg/controller/deorbit/events.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+
+	return
+}
+
+func (d *EventingDeorbiter) WaitForPersistentVolumeCleanup() (err error) {
+	d.Recorder.Eventf(d.Kluster, core_v1.EventTypeNormal, events.WaitingForDeorbitPVs, "Waiting for cleanup of Cinder volumes")
+
+	err = d.WaitForPersistentVolumeCleanup()
```
databus23: The Deorbiter interface methods for the `EventingDeorbiter` seem to call themselves. This looks like an infinite recursion to me.
Shouldn't this be something like `d.Deorbiter.WaitForPersistentVolumeCleanup()`

pkg/controller/deorbit/controller.go
```diff
@@ -0,0 +1,163 @@
...
+	return false, nil
+}
+
+func (d *DeorbitReconciler) deorbit(kluster *v1.Kluster) (err error) {
+	// The following channel is used to abort the deorbiting after a certain
+	// time. It is required because the deorbit wait-functions block indefinetly or
+	// until the stop channel is closed. This is used to unblock the workqueue.
+	var once sync.Once
+	done := make(chan struct{})
+	defer once.Do(func() { close(done) })
```
databus23: Closing a channel is not necessary for the gc to collect it, so closing it here it not necessary. That means we can remove the whole `sync.Once` logic here.

BugRoger: Removed the `sync` logic completely

pkg/controller/deorbit/deorbiter.go
BugRoger: Yes, copy-paste foo. I'm for sure going to test it all again before merge.

pkg/controller/deorbit/events.go
BugRoger: You are right.. ~Stupid auto-complete~. My mistake.

pkg/controller/deorbit/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (d *InstrumentingDeorbiter) DeleteServices() (deleted []core_v1.Service, err error) {
+	defer d.instrument("DeleteServices", time.Now(), err)
+	return d.Deorbiter.DeleteServices()
+}
+
+func (d *InstrumentingDeorbiter) WaitForPersistentVolumeCleanup() (err error) {
+	defer d.instrument("WaitForPersistentVolumeCleanup", time.Now(), err)
+	return d.WaitForPersistentVolumeCleanup()
```
databus23: Should be`d.Deorbiter.WaitForPersistentVolumeCleanup()`

BugRoger: eagle eye. fixed

PR #145, opened 2018-01-02
Adds Automatic CHANGELOG.md Generation

BugRoger:
This commits adds a Docker image that includes a tool that automatically
generates changelogs.

See: https://github.com/skywinder/github-changelog-generator

This tool relies on Github metadata to generate the Changelog. This has
the advantage that it knows more than the Git history e.g. PRs,
Labels...

The here proposed workflow is to not include the generation of the
changelog into CI, but rather manually run it for each major version.
It's easy enough:

```
  rm CHANGELOG.md
  make CHANGELOG.md
```


BugRoger:
Needs smarter Github token handling... The token has ben automatically detected and revoked...

databus23:
Nice!

But this would also require us to do manual version bumping and decide to to manage versions going forward, correct?

BugRoger:
Yes, I was thinking about aligning a minor version bump with the monthly sprints. 

The changelog can pick up all tags that fit a specific regex and group by that.

PR #155, opened 2018-01-10
Use Finalizers for Orchestrating Kluster Deletion

BugRoger:
This refactors the orchestration for Kluster Deletion. It makes use of
the Finalizers concept. It allows to shift the responsibility of
cleaning up to individual controllers. It spares us the implementation
of a more complicated kluster-phase state machine.

See: https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#finalizers

We need this as a prerequisite for the implementation of the Deorbiter
in #144.

### DIFF
--- a/pkg/controller/ground.go
+++ b/pkg/controller/ground.go
@@ -38,6 +38,8 @@ const (
 	//Reason constants for the event recorder
 	ConfigurationError = "ConfigurationError"
 	FailedCreate       = "FailedCreate"
+
+	GroundctlFinalizer = "groundctl"
 )
 
 type GroundControl struct {
@@ -191,7 +193,6 @@ func (op *GroundControl) handler(key string) error {
 					"kluster", kluster.GetName(),
 					"project", kluster.Account(),
 					"phase", kluster.Status.Phase)
-
 				if err := op.createKluster(kluster); err != nil {
 					op.Recorder.Eventf(kluster, api_v1.EventTypeWarning, FailedCreate, "Failed to create cluster: %s", err)
 					return err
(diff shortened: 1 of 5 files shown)


BugRoger:
This is fully tested and works.

pkg/controller/ground.go
```diff
@@ -432,6 +451,16 @@ func (op *GroundControl) terminateKluster(kluster *v1.Kluster) error {
...
+	if err := op.ensureFinalizerRemoved(kluster); err != nil {
+		return err
+	}
+
+	// There's a bug in the garbage-collector regarding CRDs in 1.7. It will not delete
+	// the CRD even though all Finalizers are gone. As a workaround, here we try to just
+	// delte the kluster again.
+	//
+	// This can be removed once the control-planes include garbage collector fixes
+	// for CDRs (1.8+)
```
databus23: Can you maybe provide a link to the relevant PR fixing this bug in this comment?

pkg/controller/ground.go
```diff
@@ -675,3 +704,21 @@ func (op *GroundControl) podUpdate(cur, old interface{}) {
 		}
 	}
 }
+
+func (op *GroundControl) ensureFinalizerCreated(kluster *v1.Kluster) (err error) {
```
databus23: Maybe we should move this to a util function? This will be useful in every controller and its already duplicated in this PR for `launchctl`

pkg/controller/ground.go
```diff
@@ -675,3 +704,21 @@ func (op *GroundControl) podUpdate(cur, old interface{}) {
...
+	if kluster.NeedsFinalizer(GroundctlFinalizer) {
+		return util.UpdateKluster(op.Clients.Kubernikus.Kubernikus(), kluster, func(copy *v1.Kluster) {
+			copy.AddFinalizer(GroundctlFinalizer)
+		})
+	}
+	return nil
+}
+
+func (op *GroundControl) ensureFinalizerRemoved(kluster *v1.Kluster) (err error) {
```
databus23: Maybe we should move this to a util function? This will be useful in every controller and its already duplicated in this PR for `launchctl`

pkg/controller/ground.go
BugRoger: https://github.com/kubernetes/kubernetes/issues/50528

It's not quite clear where this is fixed. This explodes into various issues and fixes. They talk about that this is back-ported to 1.7 as well. It definitely is broken in our 1.7 Kubernikus control-planes.

pkg/controller/ground.go
BugRoger: done

pkg/controller/ground.go
BugRoger: and done.

PR #160, opened 2018-01-11
Surface nova errors message for bad request responses (400)

databus23:
gophercloud at the moment just returns a generic “bad request” error message when nova returns a status 400 response for a server create request.
But the response actually contains a detailed error message as to why the payload wasn’t accepted. (e.g. given image/flavor/key not found etc.)

There is a 1 year old upstream PR solving the same thing but unfortuantely its going nowhere: https://github.com/gophercloud/gophercloud/pull/303

### DIFF
--- a/pkg/client/openstack/compute/create.go
+++ b/pkg/client/openstack/compute/create.go
@@ -26,7 +26,8 @@ func (ce *createError) Error400(e gophercloud.ErrUnexpectedResponseCode) error {
 }
 
 func (ce createError) Error() string {
-	return "Failed to create server. This should not never be printed."
+	//Unused, but we need to satisfy the error interface
+	return "Failed to create server. This shouldn't be returned ever."
 }
 
 // Create requests a server to be provisioned to the user in the current tenant.
--- a/pkg/client/openstack/compute/create.go
+++ b/pkg/client/openstack/compute/create.go
@@ -1,0 +1,43 @@
+package compute
+
+import (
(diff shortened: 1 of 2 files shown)


pkg/client/openstack/compute/create.go
```diff
@@ -0,0 +1,43 @@
...
+		} `json:"badRequest"`
+	}
+	if err := json.Unmarshal(e.Body, &response); err != nil || response.BadRequest.Message == "" {
+		return fmt.Errorf("Failed to parse response: %s", string(e.Body))
+	}
+	return fmt.Errorf("Response from compute: %d %s", e.Actual, response.BadRequest.Message)
+}
+
+func (ce createError) Error() string {
+	return "Failed to create server. This should not never be printed."
```
BugRoger: not never == always?

databus23: The deserialisation has to be defined for each error code. If quota errors result in a 400 bad request its already covered. If quota errors return something different we need to implement an additional interface for it.

databus23: Ups. You got me. this doesn't make sense. not negation to much.

databus23:
So it seems quota errors are returned as `403 Forbidden`.At the moment gophercloud doesn't support handling 403 errors with custom logic. :(

I opened an issue at gophercloud to get some feedback: https://github.com/gophercloud/gophercloud/issues/713

PR #162, opened 2018-01-11
Generate a random ssh password per kluster

databus23:
Instead of hardcoding a global password into the ignition template we now generate a random 12 char password for each kluster and use that in the ignition templates for the kluster nodes.
Existing klusters that don’t have this are staying on the old hard coded value for now.

This password is not exposed anyware at the moment. Its meant as a diagnostic backdoor during the techpreview. We can think about exposing this password to the enduser in the future.

Manually tested that this actually works.

### DIFF
--- a/charts/kube-master/templates/secrets.yaml
+++ b/charts/kube-master/templates/secrets.yaml
@@ -18,9 +18,9 @@ data:
   openstack-router-id: {{ required "missing openstack-router-id" .Values.openstack.routerID | b64enc }}
   openstack-project-id: {{ required "missing openstack-project-id" .Values.openstack.projectID | b64enc }}
   openstack-region: {{ required "missing openstack-region" .Values.openstack.region | b64enc }}
-
   bootstrapToken: {{ required "missing boostrapToken" .Values.bootstrapToken | b64enc }}
   token.csv: {{ include (print $.Template.BasePath "/_token.csv.tpl") . | b64enc }}
+  node-password: {{ required "missing node-password" .Values.nodePassword | b64enc }}
 
 {{- if empty .Values.certsSecretName }}
 {{- range list "apiserver-clients-ca-key.pem" "apiserver-clients-ca.pem" "apiserver-clients-cluster-admin-key.pem" "apiserver-clients-cluster-admin.pem" "apiserver-clients-system-kube-controller-manager-key.pem" "apiserver-clients-system-kube-controller-manager.pem" "apiserver-clients-system-kube-proxy-key.pem" "apiserver-clients-system-kube-proxy.pem"  "apiserver-clients-system-kube-scheduler-key.pem" "apiserver-clients-system-kube-scheduler.pem" "apiserver-clients-kubernikus-wormhole.pem" "apiserver-clients-kubernikus-wormhole-key.pem" "apiserver-nodes-ca-key.pem" "apiserver-nodes-ca.pem" "etcd-clients-apiserver-key.pem" "etcd-clients-apiserver.pem" "etcd-clients-ca-key.pem" "etcd-clients-ca.pem" "etcd-peers-ca-key.pem" "etcd-peers-ca.pem" "kubelet-clients-apiserver-key.pem" "kubelet-clients-apiserver.pem" "kubelet-clients-ca-key.pem" "kubelet-clients-ca.pem" "tls-ca-key.pem" "tls-ca.pem" "tls-apiserver.pem" "tls-apiserver-key.pem" "tls-wormhole.pem" "tls-wormhole-key.pem"}}
--- a/pkg/templates/ignition.go
+++ b/pkg/templates/ignition.go
@@ -8,10 +8,12 @@ import (
 	"text/template"
 
 	"github.com/Masterminds/sprig"
+	"github.com/aokoli/goutils"
 	"github.com/coreos/container-linux-config-transpiler/config"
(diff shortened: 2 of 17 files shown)


charts/kube-master/templates/secrets.yaml
```diff
@@ -18,9 +18,9 @@ data:
   openstack-router-id: {{ required "missing openstack-router-id" .Values.openstack.routerID | b64enc }}
   openstack-project-id: {{ required "missing openstack-project-id" .Values.openstack.projectID | b64enc }}
   openstack-region: {{ required "missing openstack-region" .Values.openstack.region | b64enc }}
-
   bootstrapToken: {{ required "missing boostrapToken" .Values.bootstrapToken | b64enc }}
   token.csv: {{ include (print $.Template.BasePath "/_token.csv.tpl") . | b64enc }}
+  node-password: {{ required "missing node-password" .Values.nodePassword | b64enc }}
```
BugRoger: Could you also update the chart version to `master-1.9.0-kubernikus.1`? Even though this is informational only, we can see why Kluster is using which chart version.

```
🐢  > helm ls                                                                                                                                                                                                                                                                                                         0 ↵
NAME                                  	REVISION	UPDATED                 	STATUS  	CHART                         	NAMESPACE
test2-8b25871959204ff1a27605b7bcf873f7	1       	Tue Jan  9 11:34:00 2018	DEPLOYED	kube-master-1.9.0-kubernikus.0	kubernikus
v6-5d725ddf97664a16b011e8a8dd75772b   	1       	Fri Jan 12 11:36:08 2018	DEPLOYED	kube-master-1.9.0-kubernikus.0	kubernikus-michi
```

pkg/templates/ignition.go
```diff
@@ -61,6 +63,26 @@ func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, secret *v1.Secret
 		return nil, err
 	}
 
+	//this is the old default for backwards comptibility with clusters that don't have a passwort generated
+	//TODO: Remove once all klusters are upgraded
+	passwordHash := "$6$rounds=1000000$aldshc,xbneroyw$I756LN/FtceE1deC2H.tGeSdeeelaaZWRwzmbEuO1SANf7ssyPjnbQjlW/FcMvWGUGrhF64tX9fK0abE/4oQ80"
```
BugRoger: wth? 🤣

pkg/templates/ignition.go
```diff
@@ -61,6 +63,26 @@ func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, secret *v1.Secret
...
+		//generate 16 byte random salt
+		salt, err := goutils.Random(16, 32, 127, true, true)
+		if err != nil {
+			return nil, fmt.Errorf("Unable to generate random salt: %s", err)
+		}
+		//We crank up the heat to 1 million rounds of hashing for this password (default 5000)
+		//Reason for this is we expose the resulting hash in the metadata service which is not very secure.
+		//It takes about 500ms on my workstation to compute this hash. So this means login to a node is also
+		// delayed for about a second which should be ok as this password is only meant as a last resort.
+		passwordHash, err = passwordCrypter.Generate(nodePassword, append([]byte("$6$rounds=1000000$"), salt...))
```
BugRoger: How long and complicated is the resulting password? It often needs to be entered in the crappy VPC console without copy-paste...

pkg/templates/node_1.7.go
```diff
@@ -6,7 +6,7 @@ var Node_1_7 = `
 passwd:
   users:
     - name:          core
-      password_hash: xyTGJkB462ewk
+      password_hash: {{ .LoginPassword }}
       ssh_authorized_keys: 
         - "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAvFapuevZeHFpFn438XMjvEQYd0wt7+tzUdAkMiSd007Tx1h79Xm9ZziDDUe4W6meinVOq93MAS/ER27hoVWGo2H/vn/Cz5M8xr2j5rQODnrF3RmfrJTbZAWaDN0JTq2lFjmCHhZJNhr+VQP1uw4z2ofMBP6MLybnLmm9ukzxFYZqCCyfEEUTCMA9SWywtTpGQp8VLM4INCxzBSCuyt3SO6PBvJSo4HoKg/sLvmRwpCVZth48PI0EUbJ72wp88Cw3bv8CLce2TOkLMwkE6NRN55w2aOyqP1G3vixHa6YcVaLlkQhJoJsBwE3rX5603y2KjOhMomqHfXxXn/3GKTWlsQ== michael.j.schmidt@gmail.com"
```
BugRoger: Remove this key

charts/kube-master/templates/secrets.yaml
databus23: I can do that, yes

pkg/templates/ignition.go
databus23: Thats just the old password  salted and hashed 1 million times with sha512. Nothing special.
image

pkg/templates/ignition.go
databus23: Its 12 chars and only characters and numbers. Its ok, I tested it.

pkg/templates/node_1.7.go
databus23: I will with the ssh public key PR I'm working on.

PR #163, opened 2018-01-11
Upgrades to client-go v6.0.0

### LINKED ISSUE #154: Upgrade Code-Generators
Upgrade the Go dependency to https://github.com/kubernetes/code-generator

This allows us to use the `deepcopy` generator, that will simplify the ugly copy-by-remote-call:

https://github.com/sapcc/kubernikus/blob/65ff9d8a021fdc84625fd968eee2c56af2cab061/pkg/controller/launch/pool_manager.go#L85-L87

To:

```
copy := kluster.DeepCopy()
```

It will also allow us to use an inbuilt script and get rid of the whole code-gen Makefile magic:

https://github.com/kubernetes/sample-controller/blob/258eead08702028194ade1c0a7f958c837d6f081/hack/update-codegen.sh#L28-L31

Open question: The code-generators need to be in sync with the api-machinery version. Does the api-machinery library need to be in sync with the cluster it's running on?

BugRoger:
Fixes #154


databus23:
wooho! 100k LOC removed. I'm already liking this...

BugRoger:
Github actually nicely sorts (most) of my changes to the top, if anyone wants to take a look.

databus23:
With the new `FilteredInformers` we should be able to get rid of this: https://github.com/sapcc/kubernikus/blob/client_v6/pkg/controller/operator.go#L216 yes?

PR #167, opened 2018-01-12
Specify security group by name

### LINKED ISSUE #152: Non-Default Security Group is Ignored
With the new advanced options UI/API it is now possible to select a security group. The selected security group is being ignored though.

The UI posts the correct security group id. Here I selected the security group `special_sauce` with uuid `be89fe3a-cf7e-4f8c-acd7-3871e7681d06`:

image
 
That uuid is also persisted in the Kluster spec. The created nodes are in the `default` security group though:

image

databus23:
Turns out specifying security groups by id doesn’t not work at all.
We have to specify it by name, as the official documentation demands it.

Fixes #152


pkg/client/openstack/compute/create_options.go
```diff
@@ -1,32 +0,0 @@
...
-		return nil, errors.New("Expected field `server` not found")
-	}
-
-	serverData, ok := data["server"].(map[string]interface{})
-	if !ok {
-		return nil, errors.New("Field `server` not of expected type")
-	}
-	securityGroups := make([]map[string]interface{}, len(opts.SecurityGroups))
-	for i, groupID := range opts.SecurityGroups {
-		securityGroups[i] = map[string]interface{}{"id": groupID}
```
BugRoger: Is this a "vendored" bugfix for Gophercloud?

databus23: This was my contrived way overriding gophercloud default behaviour expecting security group names (which turns out to me the only supported way). I just removed my override to change that into an `id` field.

databus23: So not its not a bugfix its just crap I wrote that we remove again.

databus23:
@edda FYI: The spec field changes from `securityGroupID` to `securityGroupName` as soon as this is merged.

databus23:
@edda Good to merge? Will this break the UI?

edda:
@databus23 Merge away. The UI is ready for the change.

Issue #153, opened 2018-01-10
Reliable Availability Alert for API


BugRoger:
We need an outside test/alert for API availability. At the moment we are blind to total outage like:

![image](https://user-images.githubusercontent.com/147369/34764241-296cda30-f5ee-11e7-922c-42cf55bdb18c.png)

auhlig:
Prometheus in the admin cluster could monitor the ingress pointing to the kubernikus api. https://prometheus.io/docs/prometheus/latest/configuration/configuration/#ingress

vonneudeck:
Also we could use the same service discovery to query a blackbox exporter outside the cluster from the prometheus within the cluster. With a bit of relabeling source and target labels we can do this with automated ingress targets.

vonneudeck:
For that we would need of course a separate blackbox-exporter that will be queried by the internal prometheus to query then the ingress.

auhlig:
This was done for Kubernikus in the above linked PR as seen by the blackbox exporter from within the kubernikus control plane.
We're working on cross regional monitoring using the exporter. Would also be deployed in similar fashion to baremetal clusters via our helm-charts.
Visualizing this in Grafana -especial cross regional overview- could be done using a combination of the following dashboards:
(1) https://grafana.com/plugins/raintank-worldping-app -> overview
(2) https://github.com/grafana/worldmap-panel -> overview
(3) table with details

vonneudeck:
4. Table with cluster (columns) and blackbox exporter (rows) location. Fields coloured gradually between green and red relative to a percentage deviation from an ideal value or median over a longer time. Could simultaneously show the absolute value in ms. 
-> quick overview and some details.

5. Variant of 2. One map results per node-exporter and one per cluster.
-> geographic overview but less details than 4.

6. Map showing a coloured connection between datacentres. Unfortunately no existing plugin seems to really support this.

BugRoger:
The internal blackbox exporters didn't catch a downtime last Saturday. The API was available but was not reachable from the outside. The internal blackbox exporters didn't catch it. 

Now that we have external blackbox probers let's add them to the monitoring setup. Make it configurable so that other's outside of our cloud can do the same.

BugRoger:
I also think we can close this one. Looks like we probe the API with an external prober now. And we do have an alert for it.

![image](https://user-images.githubusercontent.com/147369/39009025-bfae8b32-440a-11e8-8f34-b6c51d31270d.png)

[...]

PR #170, opened 2018-01-15
Implements Blackbox Monitoring for API availability

auhlig:
For #153. As discussed on Friday: This allows us to monitor availability of kubernikus API.
Blackbox exporter would come in separate PR.
TODO: 

- [x] API needs to expose at least one metric, that can be scraped. 
- [x] Alert


auhlig:
uses operator metrics for now. this would also alert if the operator is down, so we rather want the API to expose metrics than the way it is now. will do tomorrow if you agree @BugRoger.

auhlig:
exposes basic prometheus metrics and `kubernikus_api_healthz`, which is used to determine API availability

auhlig:
resolved conflicts. @vonneudeck and I are working on this. current state:
prometheus discovers ingresses via kubernetes api and sends these as targets to the blackbox exporter, which probes (GET, 2xx is ok) them.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -241,7 +241,36 @@ scrape_configs:
       target_label: __metrics_path__
       replacement: /api/v1/nodes/${1}:4194/proxy/metrics
 
-# Static Targets
+- job_name: 'blackbox-ingress'
+  metrics_path: /probe
+  params:
+    module: [http_2xx]  # Look for a HTTP 200 response.
```
auhlig: only `2xx` responses are ok for now

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -241,7 +241,36 @@ scrape_configs:
...
+  kubernetes_sd_configs:
+  - role: ingress
+  relabel_configs:
+  - source_labels: [__meta_kubernetes_ingress_annotation_prometheus_io_probe]
+    action: keep
+    regex: true
+  - source_labels: [__meta_kubernetes_ingress_path]
+    regex: /prometheus
+    action: drop
```
auhlig: prometheus won't check his-/her-/itself

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
databus23: I think this will probably not work for these https://github.com/sapcc/kubernikus/blob/master/charts/kube-master/templates/ingress.yaml

We should maybe also have a module that additionally accepts 40[13]. Can we maybe have this overridable by some annotation on the individual ingress spec? So that `http_2xx` is the default but  you can override that by specifying the module in an annotation.

charts/kubernikus/templates/ingress.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   name: kubernikus-api
   annotations:
     vice-president: "true"
+    prometheus.io/scrape: "true"
+    prometheus.io/probe: "true"
```
auhlig: need to add the `prometheus.io/probe: "true"` to consider the ingress for blackbox testing.
note to self: additional `prometheus.io/scrape: "true"` shouldn't be necessary

auhlig:
LGTM

charts/kubernikus/templates/ingress.yaml
vonneudeck: To further clarify:
These annotations are telling prometheus to do two *different* things.
`prometheus.io/probe: true` is telling prometheus to give this ingress to the blackbox exporter to test for http metrics (and return the result to prometheus).
`prometheus.io/scrape: true` is telling prometheus to scrape the ingresses targets itself for built-in prometheus metrics. Which is usually already done on pod level.

auhlig: agreed. that's why I left the note.

charts/kubernikus-system/charts/blackbox-exporter/templates/configmap.yaml
```diff
@@ -0,0 +1,50 @@
...
+        tcp:
+          query_response:
+          - expect: ^SSH-2.0-
+      tcp_connect:
+        prober: tcp
+
+      http_4xx:
+        prober: http
+        http:
+          valid_status_codes: [401,403]
```
auhlig: only `401,403` are result in `probe_success = 1`

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -244,16 +244,24 @@ scrape_configs:
 - job_name: 'blackbox-ingress'
   metrics_path: /probe
   params:
-    module: [http_2xx]  # Look for a HTTP 200 response.
+    # Look for a HTTP 200 response per default.
+    # Can be overwritten by annotating the ingress resource with the expected return codes, e.g. `prometheus.io/probe_code: "4xx"`
```
auhlig: module annotation on ingress e.g. `prometheus.io/probe_code: "4xx"`. `2xx` is default.

auhlig:
using the annotation `prometheus.io/probe_code` on an ingress will overwrite the default 2xx module, which is used in the blackbox exporter.
current limitation: if an ingress defines multiple path', the same module is used for each and every one

charts/kubernikus-system/charts/blackbox-exporter/templates/configmap.yaml
BugRoger: What is the reasoning?

charts/kubernikus-system/charts/blackbox-exporter/values.yaml
```diff
@@ -0,0 +1,13 @@
+# Default values for blackbox-exporter.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+replicaCount: 1
```
BugRoger: Would it make sense to have 2? Or is that a waste of resources...

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -241,7 +241,44 @@ scrape_configs:
...
+  # don't scrape yourself
+  - source_labels: [__meta_kubernetes_ingress_path]
+    regex: /prometheus
+    action: drop
+  - source_labels: [__meta_kubernetes_ingress_annotation_prometheus_io_probe]
+    action: keep
+    regex: true
+  # consider prometheus.io/probe_code annotation. mind below regex.
+  - source_labels: [__meta_kubernetes_ingress_annotation_prometheus_io_probe_code]
+    regex: ^(\d).+
```
BugRoger: I'm a bit confused about `http_2xx` and this code. Isn't this the expected code, that the proper should return? Or is this just for selecting the http-module?

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -241,7 +241,44 @@ scrape_configs:
       target_label: __metrics_path__
       replacement: /api/v1/nodes/${1}:4194/proxy/metrics
 
-# Static Targets
+- job_name: 'blackbox-ingress'
+  metrics_path: /probe
+  params:
+    # Look for a HTTP 200 response per default.
+    # Can be overwritten by annotating the ingress resource with the expected return codes, e.g. `prometheus.io/probe_code: "4xx"`
+    module: [http_2xx]
```
BugRoger: So, this is a default in case the annotation hasn't been overwritten?

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -241,7 +241,44 @@ scrape_configs:
...
+    replacement: http_${1}xx
+    target_label: __param_module
+  - source_labels: [__meta_kubernetes_ingress_scheme,__address__,__meta_kubernetes_ingress_path]
+    regex: (.+);(.+);(.+)
+    replacement: ${1}://${2}${3}
+    target_label: __param_target
+  - target_label: __address__
+    replacement: blackbox-exporter.kubernikus-system.svc:9115
+  - source_labels: [__param_target]
+    target_label: instance
```
BugRoger: It is my understanding that `instance` is a "reserved" label that get's filled by Prometheus. It is the instance where the metric emitted from. That would usually be the PodIP of the blackbox-exporter. 

If we had two black-box exporter pods, we couldn't distinguish between them. I'm unsure but wouldn't that even cause data "corruption" with both exporters reporting different values?

auhlig: We're using the ingress SD in this job, so the `instance` label is used for the discovered ingress and not for the blackbox exporter. The exporter itself gets discovered as a pod, where the `instance` label gets filled by prometheus

charts/kubernikus-system/charts/blackbox-exporter/values.yaml
auhlig: In the end we would have 2. 1 in the kluster, as part of kubernikus-system, and another one calling from the outside, from the admin cluster.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
auhlig: Yes. Per default a good response is `2xx`. Anything else has to be specified via the annotation on the ingress

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
BugRoger: Ah, ok. Makes sense.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
auhlig: Both. You specify the module, like `http_2xx` and that defines which response codes are accepted. The blackbox exporter receives that as a query parameter.

PR #171, opened 2018-01-15
Use filtered informers scoped to given namespace

### LINKED ISSUE #169: Migrate to FilteredSharedInformer
With `client-go` v6.0.0 there's now a `FilteredSharedInformer` that can be scoped to a namespace. This allows us to refactor our own custom filter. There's then also potential for simplifying the controller initialisation. 

As seen here: https://github.com/sapcc/kubernikus/blob/master/pkg/controller/operator.go#L216-L246

Production Example:
https://github.com/jetstack/cert-manager/blob/b978faa28c9f0fb0414b5d7293fab7bde65bde76/cmd/controller/app/controller.go#L123

See also: http://blog.kubernetes.io/2018/01/introducing-client-go-version-6.html

databus23:
To make this work we can’t use the custom “kluster” index for the pod informer anymore. We now use the standard PodLister with a label selector instead. This might incour a performance cost but I think its negligible and we only do this in state creating anyway.

Manually tested to work.

Fixes #169


BugRoger:
Nice! Hurray for negative lines of code 😄

PR #180, opened 2018-01-16
scrape kubelets, cadvisor

auhlig:
prom scrapes kubelets, cadvisor metrics. tested in k-staging.

### DIFF
--- a/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
+++ b/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
@@ -228,7 +228,6 @@ scrape_configs:
   scheme: https
   tls_config:
     ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-    insecure_skip_verify: true
   bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
   kubernetes_sd_configs:
     - role: node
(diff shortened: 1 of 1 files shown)


auhlig:
not sure about the `container_*` metrics. it's a lot.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -202,6 +202,46 @@ scrape_configs:
...
+    action: replace
+    target_label: __scheme__
+    regex: https
+    replacement: http
+
+- job_name: 'kubernetes-cadvisors'
+  scheme: https
+  tls_config:
+    ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+    insecure_skip_verify: true
```
databus23: why? Can we get this fixed?

databus23: Seems like we don't have `kubernetes.default.svc` in the added to the apiservers SANs. `kubernetes.default` should work.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -202,8 +202,46 @@ scrape_configs:
...
+  - source_labels: [__address__]
+    action: replace
+    target_label: __address__
+    regex: ([^:;]+):(\d+)
+    replacement: ${1}:10255
+  - source_labels: [__scheme__]
+    action: replace
+    target_label: __scheme__
+    regex: https
+    replacement: http
```
auhlig: this is still necessary to avoid ```Get https://198.18.0.22:10250/metrics: x509: cannot validate certificate for 198.18.0.22 because it doesn't contain any IP SANs```

auhlig:
It's ready for another review @databus23

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
BugRoger: Ok, we can't add the whole node subnet as IPs to the certificates. So, here we need to grab the unsecured port or skip host validation

Issue #97, opened 2017-11-14
Kluster migration


databus23:
Rather sooner then later we need a facility to upgrade klusters.
There are three parts to kluster upgrades:

1. spec changes --> `helm upgrade`
2. crd/secret changes --> a to be written/discovered migration system
3. node updates --> recreate

For 2. maybe https://github.com/Boostport/migration could be something. It supports generic migrations expressed in golang.

An open question would be would which of these steps is user controller and which happens automatically:

1. Probably the user
2. Is the tricky one, some migrations are mandatory and should be automatic when a new version of the operator/apiserver is deployed. Other might be depended on an `helm upgrade`. :|
3. Most definitely the user

PR #182, opened 2018-01-17
Add kluster spec migration

databus23:
This is an attempt provide a fix for #97.
Specifically this PR provides a lightweight way of automatically migrating kluster specs to newer versions (e.g. add/rename fields with auto discovered values etc).
This is **not** intended to facilitate upgrades of klusters (`helm upgrades`) but update our internal representation of a kluster.

A little explanation:
* The ordered list of migration steps is kept in a go slice in `pkg/migration/register.go`. The versioning is based on simply enumerating the migrations steps (4 migrations steps means the latest version is 4) 
* The spec/schema version of a Kluster is kept in `Status.SpecVersion`. New Klusters are created with the latest Version of the apiserver.
* A migration step is defined by a reasonable simple function signature (see `migration.Migration`)
* During operator startup before the controllers are started all existing klusters pending migrations are upgraded. I did it explicitly this way to avoid loosing the raw Data from a previous spec (e.g. a field was removed). Because as soon as a controller updates a kluster the internal json representation is completely replaced.

Open questions:
1. When upgrading a kluster fails it is not retried until the operator is restarted. Do we want to have a controller type logic that reconciles this continuously? If so we need to also think if we want to have controllers ignore klusters that are pending migrations. Otherwise we risk having them operate on incomplete data and destroy old information that would maybe be required for completing the migration.
2. Should be maybe call it `SchemaVersion` instead of `SpecVersion` to avoid confusion?


SchwarzM:
Hi,

i am currently missing code that forbids a version 3 operator to handle version 4 klusters.
In general I think during a rolling update we will have version 3 and version 4 operators in parallel.
If I would inject a version 1 kluster (with kubernikusctl) both would try to migrate it.

Kind regards
M. Schwarz

BugRoger:
As discussed:

- Migrations run before the operators are started
- If a migration fails, the kluster will change phase to `disabled`
- Controllers are supposed to ignore `disabled` klusters
- Add a Migration controller that periodically tries to migrate `disabled` klusters and emit events
- Change phase updates to `patch` in order to keep outdated spec values
- Please also add a little bit of documentation to `docs/development/migrations.md` explaining how this works and what the expectations are

databus23:
On second thought introducing a `Disabled` phase is not a good idea as we will have the same problems we had with the `Error` phase: We don't know which phase the Kluster was in before it got disabled. I think we need to express this as some kind of `Condition`. Maybe as simple as a boolean `Status.Disabled`

databus23:
This is ready for another review round. I added a small amount of documentation explaining the basic.
Here are the basics:
* I added a *Condition* `Status.MigrationsPending`, that is updated before controllers start.
* Controllers that modify the Kluster are supposed to check `Disabled()` on a `Kluster` resource and exit early. The idea is that in the future other conditions might disable a kluster and controller don't need to know the specifics just that they need to ignore this Kluster
* The migration controller reconciles `Status.MigrationStatus` for all klusters continuously (tries to migrate klusters that have migrations pending).

databus23:
@BugRoger Care ro review? I consider this ready fo review.

docs/development/migrations.md
```diff
@@ -0,0 +1,22 @@
...
+==========
+Kubernikus incorporates a simple migration system to update the internal representation of existing Klusters.
+This can be used to update the CRD of a Kluster or update any other accompanying kubernetes resource in the control plane (e.g. Kluster Secret, Ingress, etc).
+
+To create a migration a [Migration](https://github.com/sapcc/kubernikus/blob/master/pkg/migration/migration.go) function needs to be written and [registered](https://github.com/sapcc/kubernikus/blob/master/pkg/migration/register.go).
+
+```
+type Migration func(klusterRaw []byte, kluster *v1.Kluster, client kubernetes.Interface) (err error)
+```
+* `klusterRaw` contains the current raw JSON data of the Kluster, it can be used to access old data not exposed by the `v1.Kluster` type anymore.
```
BugRoger: I have a vague recollection that we discussed about this, but why is `klusterRaw` a `json` blob? Wouldn't it be easier to access if it were a deep-copy of `kluster`?

docs/development/migrations.md
```diff
@@ -0,0 +1,22 @@
...
+* `client` is a kubernetes client which can be used to modify other kubernetes resources belonging to the Kluster.
+
+The migration controller takes care of invoking the defined function for every Kluster until it succeeds. When a migration is successfully executed the `Status.SpecVersion` is updated to reflect the migration has been applied to the respective Kluster.
+
+### Pending migrations
+As long as a kluster is not up to date with respect to `Status.SpecVersion` it will have the condition `Status.MigrationsPending` set to true.
+Klusters with pending migrations are ignored by all controllers that modify the internal Kluster representation (e.g. update the CRD).
+`Status.MigrationsPending` is updated during operator startup before any controller is running and continuously kept in sync by the migration controller.
+
+**Controllers should call the `Disabled()` method on the `v1.Kluster` and exit early in reconciliation functions in case a Kluster is disabled.**
```
BugRoger: Sounds like we have to add this into the other existing controller loops once this is merged.

BugRoger: Ah, nvm.

pkg/controller/migration/controller.go
```diff
@@ -0,0 +1,65 @@
...
+
+	var reconciler base.Reconciler
+	reconciler = &MigrationReconciler{clients, recorder, logger}
+
+	return base.NewController(threadiness, factories, reconciler, logger, nil)
+}
+
+func (mr *MigrationReconciler) Reconcile(kluster *v1.Kluster) (bool, error) {
+
+	//We only care about klustes with pending migrations
```
BugRoger: kluster

pkg/controller/migration/controller.go
```diff
@@ -0,0 +1,65 @@
...
+	if !migration.MigrationsPending(kluster) {
```
BugRoger: These too loops are a not obvious. Double negation and noop in the called function. Feels like it could be unraveled.

pkg/controller/operator.go
```diff
@@ -185,8 +188,11 @@ func (o *KubernikusOperator) Run(stopCh <-chan struct{}, wg *sync.WaitGroup) {
 	o.Factories.Kubernikus.WaitForCacheSync(stopCh)
 	o.Factories.Kubernetes.WaitForCacheSync(stopCh)
 
-	o.Logger.Log("msg", "Cache primed. Ready for Action!")
-
+	o.Logger.Log("msg", "Cache primed")
+	if _, found := o.Config.Kubernikus.Controllers["migration"]; found {
```
BugRoger: Is this really necessary here? Wouldn't the immediate first reconciliation be equivalent?

pkg/migration/migration.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return nil
+}
+
+func migrateKluster(kluster *v1.Kluster, version int, migration Migration, client kubernetes.Interface, kubernikus_client kubernikus.Interface) (*v1.Kluster, error) {
+	var rawData []byte
+	var err error
+
+	//TODO: Don't import fake pkg outside of test code
+	if _, ok := kubernikus_client.(*kubernikusfake.Clientset); !ok {
```
BugRoger: I suppose the FakeClient doesn't mock the `RESTClient()`?

pkg/migration/migration.go
```diff
@@ -0,0 +1,90 @@
...
+	return defaultRegistry.Migrate(k, client, kubernikus_client)
+}
+
+//Registry manages an ordered list of migration steps
+type Registry struct {
+	migrations []Migration
+}
+
+//AddMigration appends a migration to the list
+func (r *Registry) AddMigration(m Migration) {
```
BugRoger: Hm. Coming from Rails where the migrations are ordered by a unique date prefix, I was confused how this works here. I think, this `Registry` assumes that the migrations are always added at the end of the `register` function. And that there is no change in order or deletions.

I think, it's cool to keep it simple. But we should probably add a comment into the register call to not change it. Especially to not delete previously commited migrations.

pkg/migration/register.go
```diff
@@ -0,0 +1,7 @@
+package migration
+
+func init() {
```
BugRoger: Add safeguard comment to not mess around with existing migrations.

docs/development/migrations.md
databus23: The problem is that the migration runs in the new codebase with the new `CRD` model. That means if you removed a field (or renamed, changed the structure) you loose the old data. So loading it into a `Kluster` is not an option.

pkg/migration/migration.go
databus23: Spot on.

pkg/controller/operator.go
databus23: The idea was to have the klusters migrated before any controllers are running to avoid running controllers that expect the new model on an old model that hasn't been migrated yet.
Also to not accidentally modify models that haven't been migrated yet (so that we don't loose the old data we might need for the migration) we need to make sure they are updated to be `Disabled()` before other controllers run.

pkg/migration/migration.go
databus23: You are absolutely right with your assessments. I'll add the comment and would not invest more time at this point to make it more like a full-blown migration system (e.g. as in rails).

pkg/migration/register.go
databus23: done

pkg/controller/migration/controller.go
databus23: I'm not sure I follow. Care to elaborate?

pkg/controller/migration/controller.go
databus23: Fixed, thanks.

pkg/controller/operator.go
BugRoger: Ah, ok. Good. Carry on... :)

pkg/migration/migration.go
BugRoger: Cool!

PR #184, opened 2018-01-18
Fix shutdown of operator metrics exporter

databus23:
The metrics exporter was not shutting down because it was blocked in `http.ListenAndServe`. @auhlig Kind of the same misconception as we found in the #121 PR


auhlig:
lgtm

BugRoger:
lgtm

PR #186, opened 2018-01-18
fix doc images

jonathanschwarze:
Change path for images to make it accessible in Hugo. Furthermore, delete slack contact section.


notque:
Looks like i missed a slack section, thanks for grabbing it!

PR #187, opened 2018-01-19
Add routegc controller

databus23:
This controller starts a watch loop for every kluster monitoring the kluster’s router.

It automatically removes routes that reside within the `ClusterCIDR` and point to an non address that can’t be matched on an exiting instance in nova.

It is a mitigation for kubernetes/kubernetes#56258 which fixes this problem upstream for k8s 1.10+

Closes #116 (actually it replaces it)

*Its kind of unrelated but snuck in this PR:* I refactored the `Controller` interface to not require a `threadiness` argument because for this controller it doesn't apply and it shows that there are controllers (e.g. polling ones like this one) where `threadiness` doesn't apply. I think it reduces the `Controller` interface to the essentials and everything else is left to the controllers individual `New` method.


pkg/client/kubernetes/tunnel.go
```diff
@@ -59,7 +59,6 @@ func NewTunnel(client rest.Interface, config *rest.Config, namespace, podName st
 // Close disconnects a tunnel connection
 func (t *Tunnel) Close() {
 	close(t.stopChan)
-	close(t.readyChan)
```
BugRoger: Is this correct? That channel get's close nowhere now...

pkg/controller/base/controller.go
```diff
@@ -38,16 +38,18 @@ type controller struct {
 	queue      workqueue.RateLimitingInterface
 	reconciler Reconciler
 
-	logger log.Logger
+	logger      log.Logger
+	threadiness int
 }
 
-func NewController(factories config.Factories, clients config.Clients, reconciler Reconciler, logger log.Logger) Controller {
+func NewController(threadiness int, factories config.Factories, clients config.Clients, reconciler Reconciler, logger log.Logger) Controller {
```
BugRoger: Should this whole threadiness change be part of this PR?

pkg/controller/routegc/controller.go
```diff
@@ -105,30 +133,70 @@ func (r *RouteGarbageCollector) Reconcile() {
...
+			return
+		}
+		kluster := obj.(*v1.Kluster)
+		logger := log.With(r.logger, "kluster", kluster.Name)
+		begin := time.Now()
+		err = r.reconcile(kluster, logger)
+		logger.Log("msg", "Reconciling", "took", time.Since(begin), "v", 5, "err", err)
 	}
+	wait.JitterUntil(reconcile, r.syncPeriod, 0.5, true, stop)
+	fmt.Println("done")
```
BugRoger: Forgotten debug

pkg/client/kubernetes/tunnel.go
databus23: thats correct. While debugging I got multiple panics  because of a double close at this point. The tunnel.go method is copied from helm and that had this changed as well: https://github.com/kubernetes/helm/commit/fe3eeaf39d44a95aa9e7cff29647cd12ab60da5a.
Its totally unrelated but not worth a separate PR I think.

pkg/controller/routegc/controller.go
databus23: Ups

databus23:
That todo bot has a funny way of showing up

pkg/controller/base/controller.go
databus23: It an ideal world it should probably not. This is the first controller that requires this change though as `threadiness` doesn't apply for this controller. Instead of doing a separate PR only refactoring the controller interface I decided to shim it in here to avoid the roundtrips of 2 PRs. We can do it separately if you prefer.

databus23:
@BugRoger This is ready for another review.

pkg/client/kubernetes/tunnel.go
BugRoger: Ok, sure thing

pkg/controller/base/controller.go
BugRoger: Ok

PR #205, opened 2018-01-30
Add Monitoring Dashboards

BugRoger:
This PR adds Grafana as another `kubernikus-system` chart. It also includes a regional Kubernikus monitoring dashboard. It will be available at `/grafana`.

The application is installed stateless. All configuration is provisioned using the inbuilt mechanisms in Grafana. Dashboards need to be added to the Helm values file.

The whole alerting and user-management implementation is disabled. By default all dashboards are read-only and accessible without authentication.

There's an `admin` user which can be used to create or modify dashboards. Just make sure to export the json and add it to the configmaps. Otherwise changes will be lost once the pods restarts.

Unfortunately, the Grafana chart had to be vendored. There are a few changes required in order to enable the provisioning mechanisms in Grafana 5.0.0.


databus23:
In k-eu-nl-1:
```
kubectl get configmap kubernikus-system.v99 -o json | jq -c . | wc -c
   60757
```
60kb is the compressed release, not to shabby. Gives is enough room for the forseable future (Famous last words)

databus23:
but on second thought, this approach allow for not more than 10 reasonable complex dashboards before the configmap will become a bottleneck.

BugRoger:
> Can we file an upstream PR to get this included eventually?
Totally, I will create the PR and try to remove the vendoring completely. Let's see if I manage to pull in a forked Helm chart. I have my doubts... ;)

> the configmap will become a bottleneck.
I don't like that the dashboards need to be passed as Helm values. That pushes the problem into the deployment step (merge all dashboards into a values.yaml). The goal is to make the deployment easier though. 

The end-goal of this exercise is, that we can use the community Grafana chart without modification. Additionally, I would like to separate the dashboards completely from that chart. The dashboards should be part of the application and whatever Grafana feels responsible for them, can pick them up.

With Grafana 5 there's a new provisioner feature. Provisioners are an integral part of Grafana (no sidecar or operator) and will automatically recheck and reload dashboards. With this we can implement a ConfigMap/CRD Provisioner. 

Given that, we put the dashboards configmaps into the application's Helm charts. The provisioner will listen to updates and automatically pull in the new/updated dashboard. That completely decouples the charts and detangles the deployment of app and dashboard.

BugRoger:
PR is in https://github.com/kubernetes/charts/pull/3514/files

That allowed me to remove Grafana from this PR. It leaves us with a mechanism to allows to extend the Prometheus charts via Helm values. 

@auhlig please double check that you are cool with this. Not sure you noticed before.

PR #207, opened 2018-01-31
Use nodeobservatory to update node pool status

databus23:
This change refactors the NodeObservatory to act more like an informer instead of a controller. It also adds a factory for it so that the NodeObservatory only started when it actually request by a controller.
Secondly launchctl now uses the NodeObservatory to update the `Healthy` and `Schedulable` status of node pools.

### DIFF
--- a/pkg/controller/launch/pool_manager.go
+++ b/pkg/controller/launch/pool_manager.go
@@ -1,15 +1,21 @@
 package launch
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/sapcc/kubernikus/pkg/api/models"
 	"github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
 	"github.com/sapcc/kubernikus/pkg/client/openstack"
 	"github.com/sapcc/kubernikus/pkg/controller/config"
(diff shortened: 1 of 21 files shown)


databus23:
@auhlig We need to take a look at the test tomorrow. Its not working anymore because GetStoreForKluster does not exist anymore. I think we should try to come up with some other things to test.

auhlig:
I tested in k-staging. Current unit test is just checking for obvious errors on happy path, as the rest of the logic is quite hard to test. Can we have a call tomorrow?

pkg/controller/launch/pool_manager.go
```diff
@@ -74,12 +81,29 @@ func (cpm *ConcretePoolManager) GetStatus() (status *PoolStatus, err error) {
 }
 
 func (cpm *ConcretePoolManager) SetStatus(status *PoolStatus) error {
+	var healthy, schedulable int64
```
BugRoger: It would be cool to break out the `healthy` and `schedulable` fields into the `PoolStatus`. That would allow us to do things like self-healing and make the health status known to the controller too.

databus23:
@BugRoger I added the fields to `NodePoolStatus`

PR #208, opened 2018-02-01
fix install for kluster names longer then 11 chars

databus23:
turns out labels also are limited to 64 characters, who knew…

This is a follow up for #193

### DIFF
--- a/charts/kube-master/templates/controller-manager.yaml
+++ b/charts/kube-master/templates/controller-manager.yaml
@@ -15,11 +15,15 @@ spec:
       maxSurge: 1
   selector:
     matchLabels:
-      app: {{ include "master.fullname" . }}-controller-manager
+      app: controller-manager
+      kluster: {{ .Values.name }}
+      account: {{ .Values.account }}
   template:
     metadata:
       labels:
-        app: {{ include "master.fullname" . }}-controller-manager
+        app: controller-manager
+        kluster: {{ .Values.name }}
+        account: {{ .Values.account }}
         release: {{ .Release.Name }}
       annotations:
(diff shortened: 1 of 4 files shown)


charts/kube-master/templates/controller-manager.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: "extensions/v1beta1"
 kind: Deployment
 metadata:
-  name: {{ include "master.fullname" . }}-c-manager
+  name: {{ include "master.fullname" . }}-cmanager
```
databus23: By renaming the deployment we should handle the upgrade case better for exiting klusters. The old deployment will be deleted and a new one will be created. This should prevent the we leave orphaned replicates/pods behind as explained here https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates

databus23:
@BugRoger no because we keep all klusters in a single namespace und two projects with a kluster named *test* would not be possible. *Sorry the kluster name test is already taken* 🤣

databus23:
@bugroger we already have the projected as a label *account* btw https://github.com/sapcc/kubernikus/blob/kluster-name/pkg/api/handlers/create_cluster.go#L53

BugRoger:
We could add the usual 6 chars hex garble to get `unique` test klusters. ;)

databus23:
We had this discussion before, then its quite hard in the apiserver to know which crd to fetch for `/api/v1/clusters/:name`. How do we know which garble?
I think its fine how it is now for know. Don't bogart that joint my friend... :)

BugRoger:
I seem to have selective memory on that one. ;) Anyway approved. Fly away little bird...

Issue #220, opened 2018-02-07
Add Alert for Controller Operation Error Spikes 


BugRoger:

auhlig:
closed by #227

Issue #219, opened 2018-02-07
Add Alert for Hanging Controllers


BugRoger:
Track the rate of operations of each controller. Due to the periodic and constant reconciliations there is always something going on.

If the rate of operations drops to 0 we need to fire an alert.

auhlig:
closed by #227

Issue #218, opened 2018-02-07
Alert for Goroutine Spikes in Operator


BugRoger:
Detect Goroutine leaks.

Related to #216

PR #227, opened 2018-02-13
add alerts

auhlig:
for issues: #218 #219 #220 
Did you have something like this in mind @BugRoger ?


auhlig:
Yes. No metrics for `groundctl`.
Also deorbiter and launchctl expose `kubernikus_deorbit_operation_latency_*` metrics, which is missing for for routegc.  Also routegc does not track `kubernikus_routegc_operation_total`.

PR #228, opened 2018-02-14
WIP: Move k8sniff chart

auhlig:
The k8sniff chart is part of kubernikus-system and should reside there. 
Also moved misplaced scrape annotation from deploy to pod template.


auhlig:
Thanks. Missed that. Remove. 
Not sure why there was a dependency, since the sniffer is needed in for klusters created with the kube-master chart

BugRoger:
That `enabled` flag is also in the "secret" values at the usual place.

auhlig:
Added. I'll check again tomorrow if this is good now.

Issue #213, opened 2018-02-07
Probe Customer API Servers


BugRoger:
Probe the availability of the customers API servers using the external Blackbox exporters.

vonneudeck:
Add `prometheus.io/probe` annotation here for service discovery
https://github.com/sapcc/kubernikus/blob/f1735344d84e9fb71403d76ff5bab88d9a83ebb3/charts/kube-master/templates/ingress.yaml#L14-L22

Add job with three (or more) zones here
https://github.com/sapcc/kubernikus/blob/f1735344d84e9fb71403d76ff5bab88d9a83ebb3/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl#L254

auhlig:
No quick fix possible. I had to silence the alerts for this, as they were thrown by each and every kluster.
The [kube-master ingress](https://github.com/sapcc/kubernikus/blob/master/charts/kube-master/templates/ingress.yaml) defines 2 hosts: 1 for the API server and 1 for the wormhole. Both return different response codes in case of success.
The customer probing by @vonneudeck uses the initial prober setup, which can not handle different response codes per ingress (same module, e.g. 2xx,4xx, .. would work, but not 1 with 2xx and the other with 4xx).
Customer api probing needs fixing before the alert silence expires end of this week else please extend the silence @vonneudeck

BugRoger:
I'd say we can close this. Customer API's are being probed now \o/

PR #230, opened 2018-02-14
 Probe ingresses in kubernikus-system over region_probed_from

vonneudeck:
Solves Issue #213 

We probe all ingresses in kubernikus-system every five seconds with the
blackbox-exporter/prober from all the regions mentioned in the list
`region_probed_from`.
If the region the ingress is running in is part of the list it will
be treated specially and probe from inside the cluster using the
internal service instead of the ingress.

This was done to not be flooded in redundant probes while stile having
a probe from outside of the region whose number can be raised easily.
The rationale behind the handpicking of the regions listed in `region_probed_from`
is to be able to define a region that is nearby to probe from and to be flexible to add more.

This can also be used to probe other ingresses by adding the annotation `prometheus.io/probe: "true"`

### DIFF
--- a/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
+++ b/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
@@ -250,8 +250,8 @@ scrape_configs:
       regex: (.+)
       target_label: __metrics_path__
       replacement: /api/v1/nodes/${1}:4194/proxy/metrics
+{{- range .Values.region_probed_from }}
 
-{{- range .Values.prometheus.region_probed_from }}
 - job_name: 'blackbox-ingress-{{ . }}'
   scrape_interval: 5s
   metrics_path: /probe
--- a/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
+++ b/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
@@ -294,7 +294,7 @@ scrape_configs:
     target_label: kubernetes_name
   - source_labels: [__meta_kubernetes_ingress_path]
     target_label: path
-  - target_label: region_probed_from
+  - target_label: probed_from
     replacement: {{ . }}
 {{- end}}
 
--- a/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
+++ b/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
@@ -250,53 +250,6 @@ scrape_configs:
       regex: (.+)
       target_label: __metrics_path__
       replacement: /api/v1/nodes/${1}:4194/proxy/metrics
-{{- range .Values.region_probed_from }}
-
-- job_name: 'blackbox-ingress-{{ . }}'
-  scrape_interval: 5s
-  metrics_path: /probe
-  params:
-    # Look for a HTTP 200 response per default.
-    # Can be overwritten by annotating the ingress resource with the expected return codes, e.g. `prometheus.io/probe_code: "4xx"`
-    module: [http_2xx]
-  kubernetes_sd_configs:
-  - role: ingress
-  relabel_configs:
-  # don't scrape yourself
-  - source_labels: [__meta_kubernetes_ingress_path]
-    regex: /prometheus
-    action: drop
-  - source_labels: [__meta_kubernetes_ingress_annotation_prometheus_io_probe]
-    action: keep
-    regex: true
-  # consider prometheus.io/probe_code annotation.
-  - source_labels: [__meta_kubernetes_ingress_annotation_prometheus_io_probe_code]
-    regex: (\b\w{3,})
-    replacement: http_${1}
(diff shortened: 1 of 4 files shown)


charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -277,7 +279,11 @@ scrape_configs:
     replacement: ${1}://${2}${3}
     target_label: __param_target
   - target_label: __address__
+  {{- if eq . $.Values.global.region }}
     replacement: blackbox-exporter.kubernikus-system.svc:9115
+  {{- else }}
+    replacement: prober.{{ . }}.cloud.sap
```
auhlig: Do we want to use the prober from our bm clusters or do we want something like `prober.kubernikus.<region>.cloud.sap`?

vonneudeck: I am in favour of having as few blackbox-exporters as possible. It will get messy early enough and an overview of the infrastructure is already a bit daunting.
Right now, I would stick with the ones on the BM, especially as kubernikus is not rolled out everywhere yet.
If we want to move load away from the BM in the future, I would not change the URI in single configurations but rather let the DNS-entry for `prober.$region.cloud.sap` point somewhere else.

auhlig:
The ingress spec defines [2 rules](https://github.com/sapcc/kubernikus/blob/master/charts/kube-master/templates/ingress.yaml#L14-L29), one for the api server and one for the wormhole. We would probe both using the above configuration. Could this be an alternative solution to #114?

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
auhlig: Then it could be less daunting if we had the same setup in every cluster.

BugRoger: I would like to propose to explicitly define a (list of) proper-urls in the values of the chart. That allows us to decouple this configuration from the SAP specific setup. Additionally, we don't need to generate the proper URL here. It also opens up possibilities for completely different setups, like having SaaS hosted probers or installations in a different cloud altogether:

```
blackbox-exporters:
  - http://pingdom.com/blackbox-exporter
  - https://prober.eu-de-1.cloud.sap
  - https://proper.na-us-1.cloud.sap
  - http://blackbox-exporter.kubernikus-system.svc:9115
```

One downside of that approach is that it also requires a loop (or it will hurt with dry-coding OCD). Another idea, would be to just have:

```
blackbox-exporters:
  primary: http://pingdom.com/blackbox-exporter
  secondary: https://prober.eu-de-1.cloud.sap
```

That would allow to simplify the config and wouldn't require a golang template loop at all. With the helm double evaluation that comes with it. Extra points because it will be easy to stick it into the `extra_scrape_config` and move the whole configuration out of the Prometheus chart. That gets us one step closer to being able to use the `stable/prometheus` chart.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -288,6 +294,9 @@ scrape_configs:
     target_label: kubernetes_name
   - source_labels: [__meta_kubernetes_ingress_path]
     target_label: path
+  - target_label: region_probed_from
+    replacement: {{ . }}
+{{- end}}
```
BugRoger: Isn't that already clear from the proper URL? It's also another drop of special SAP sauce...

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -251,7 +251,9 @@ scrape_configs:
       target_label: __metrics_path__
       replacement: /api/v1/nodes/${1}:4194/proxy/metrics
 
-- job_name: 'blackbox-ingress'
+{{- range .Values.prometheus.region_probed_from }}
```
BugRoger: My personal preference is to stick this config into the `extra_scrap_config` of the values. We'd have to dumb it down a bit though. Maybe even into a single global config for all installations. Like always probe from `proper.eu-nl-1.cloud.sap` and `proper.na-us-1.cloud.sap` and that's it.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
BugRoger: I'd even be happy with a single configurable prober...

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
vonneudeck: Good point. By default the prober URI is dropped, but I will use a more generic no-SAP way to keep the information.

vonneudeck:
I changed everything to be pulled in from`extra_scrape_config`.

PR #243, opened 2018-02-23
WIP: use stable/prometheus chart

auhlig:
- use [stable/prometheus chart](https://github.com/kubernetes/charts/tree/master/stable/prometheus) instead of our custom frickelloesung
- use [fusion](https://github.com/sapcc/kubernetes-operators/tree/master/prometheus-fusion) to collect and fuse prometheus rules and alerts and update prometheus configmap
- use [promtheus sidecar](https://github.com/jimmidyson/configmap-reload) to reload configmap on change
- one blackbox exporter to probe them all: remove from kubernikus. use only bm probers.


charts/kubernikus-system/templates/_prometheus.yml.tpl
```diff
@@ -8,15 +8,13 @@ global:
   external_labels:
     region: {{ .Values.global.region }}
 
-{{- if .Values.use_alertmanager }}
```
BugRoger: What is the reason this isn't needed anymore?

charts/kubernikus-system/templates/prometheus-alerts.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: common-prometheus-rules
+  labels:
+    app: fusion
+    type: rule
```
BugRoger: Is there another type except `rules`? I suppose, the `app` label is being used for selecting on the ConfigMaps that contain Prometheus-Fusion config. Isn't this double bubble with the annotation below?

charts/kubernikus-system/templates/prometheus-alerts.yaml
```diff
@@ -0,0 +1,11 @@
...
+  annotations:
+    prometheus.io/rule: "true"
```
BugRoger: What is the annotation good for?

charts/kubernikus-system/values.yaml
```diff
@@ -12,6 +12,62 @@
...
+    minReadySeconds: 20
+    terminationGracePeriodSeconds: 300
+    strategy:
+      type: Recreate
+    persistentVolume:
+      size: 300Gi
+    resources:
+      limits:
+        cpu: 500m
+        memory: 512Mi
```
BugRoger: Seems a bit conservative. Did you check how much memory we actually need?

charts/kubernikus-system/vendor/prometheus-fusion/values.yaml
```diff
@@ -0,0 +1,36 @@
+fusion:
+  image: sapcc/prometheus-fusion
+  tag: v201804161653
+
+# enable RBAC
+rbac:
+  enabled: false
```
BugRoger: I would default to `true`

charts/kubernikus-system/templates/_prometheus.yml.tpl
auhlig: We’re always using the alertmanager. Could also leave the condition and default to `true`

charts/kubernikus-system/vendor/prometheus-fusion/values.yaml
auhlig: K

charts/kubernikus-system/templates/prometheus-alerts.yaml
auhlig: The `labels` of an alert/rule configmap would describe the app they belong to. So in Kubernetes it would be like
```
app: kubernikus
type: api
```
The above for the common rules is a bad example and should be changed to `app: Kubernetes`

charts/kubernikus-system/templates/prometheus-alerts.yaml
auhlig: The operator picks up configmaps with this annotation

charts/kubernikus-system/values.yaml
auhlig: I did. But I think just in k-feature not in a productive env. Good point.

PR #245, opened 2018-02-24
use stable/nginx-ingess

auhlig:
This is a very stabel nginx-ingress.
No reason to keep all of this in charts. Use official stable/nginx-ingress chart. - 1114 LOC.
Tested in `k-qa-de-1`.


auhlig:
Also enforced some resource limits. This is good to go, if there are no objections from your side @BugRoger, @databus23?

BugRoger:
Hard to review but it's right direction. Bring it on...

auhlig:
Okay. On 2nd though: I'll merge this Monday morning

PR #246, opened 2018-02-25
Flight Controller that Automatically Fixes Common Problems

BugRoger:
# Flight Control

This controller takes care about Kluster health. It looks for obvious problems and tries to repair them. It will:

- [x] Reconcile the security group on each instance
- [x] Use ConfigDrive instead of Metadata Agent (Implemented in #249)
- [x] Delete Incompletely Spawned Instances
- [x] Reconcile PodCIDRs added to the Kluster's Security Group

## Currently Implemented

  * It deletes Nodes that didn't manage to register within 10m after
   inital creation. This is a workaround for DHCP/DVS (latency) issues.
   In effect it will delete the incompletely spawned node and launch
   control will ramp it back up.

   * It ensures tcp/udp/icmp rules exist in the security group defined during
   kluster creation. The rules explicitly allow all pod-to-pod
   communication. This is a workaround for Neutron missing the
   side-channel security group events.

   * It ensures each Nodes is member of the security group defined in
   the kluster spec. This ensures missing security groups due to whatever
   reason are again added to the node.

 ## 1. Security Group (DVS) Update Latency

Updates to the security group are not instant. There is a non-trivial amount of latency involved. This leads to timeouts and edge cases as the definition of tolerable latency differs. Most notable this affects DHCP and Ignition. Default tolerances are in the range of 30s.

The latency of updates is related to general load on the regions, "noisy" neighbors blocking the update loops, amount of ports in the same project and Neutron and VCenter performance.

 ### Symptoms

  * Nodes Running but never become healthy
  * No IPv4 Address visible on the VNC Console
  * No SSH Login Possible
  * Kubelet not running

These symptom indicates that the node couldn't configure its network interface before the Ignition timeout. This effectifly leaves the node broken.

 ### Workarounds

  * **Increase DHCP/Ignition Timeout**
     This configuration needs to be baked into the image as an OEM customization. It also interacts with the DHCP client timeout which again requires a modification of the image. With frequent CoreOS updates this modification needs to be automatic and included in the image build pipeline.

  * **Reboot the Instance**
     ~This is the preferred workaround~. It gives the DVS agents additional time to configure the existing image and retries the Ignition run. This does not work. Ignition only runs on first boot. Failure also counts as first run. The node is broken if it doesn't work the first time.

  * **Delete the Instance**
    This workaround is problematic. It will not succeed if the update latency is too high in general.

* **Use ConfigDrive**
     It is possible to pass in metadata via config-drive. The initial configuration of a node is independent of network connectivity. With this a node is guaranteed to boot correctly. It will not be functional but at least be in a recoverable state.  

 ## 2. Security Group Update Event Missed

If an instance successfully downloads and startes the Kubelet, it registers itself with the APIServer and gets a PodCIDR range assigned. This triggers a reconfiguration of the Neutron Router and adds a static route. The route points the PodCIDR to the node's IP address. This is required to satisfy the Kubernetes pod to pod communication requirements.

As the PodCIDR subnet is now actually routed via the Neutron router it is required to be allowed in the security group.

This happens by updating the node's Neutron port and adding the required CIDR to `allowed_address_pairs`. This triggers an event that the port was updated. The DVS agent are catching this update and adding an additional rule to the security group.

Occasionally, this update is missed. Until a full reconcilation loop happens (usually by restarting or update of the DVS agents) the following symptoms appear.

### Symptoms

  * Sporadic Pod Communication
  * Sporadic Service Communication
  * Sporadic Load Balancer Commnication
  * Sporadic DNS Problems in Pods
     Depending on the disconnected node pods can't reach the Kube-DNS service. DNS will work on the nodes.
  * Load Balancer Pools Flapping

 ### Workarounds

  * **Add PodCIDRs to Security Group**
      Instead of relying on the unreliable Oslo events, all possible PodCIDRs are being added to the kluster's security group. Per default this is 198.19.0.0/16

  * **Trigger Security Group Sync**
     Instead of waiting for a security group reconcilliation force an update by periodially add a (random) change to the security group. If possible this should only be triggered when a node condition indicates pod communication problems.

 ## 3. Default Security Group Missing
 
When a new node is created via Nova a security group can be specified. The user can select this security group during Kluster creation. If nothing is selected the `default` security group is assumed.

For yet unknown reasons, there's a chance that the instance is configured without this security group association by Nova. In effect the instance is completely disconnected from the network.

### Symptoms
  
  * Nodes Running but never become healthy
  * No IPv4 Address visible on the VNC Console
  * No SSH Login Possible
  * Kubelet not running

 ### Workarounds

  * **Reconcile Security Group Associations**
     Periodically check that instances which haven't registerd as nodes do have the required security group enabled. If not, set it again.

 ## 4. ASR Route Duplicates
 
 When a node is being deleted its route is removed in Neutron. The ASR agents get notified by an event and do remove the route from the ASR device.

First of all, this requires that the state of the Neutron DB reflects reality. Updates to the routes are done by the RouteController in the Kubernetes OpenStack cloud provider. Before 1.10 there's a bug that misses the updates. In Kubernikus we fixed this by adding an additional RouteNanny for now.

When a Kluster is terminated forcefully, the RouteController might be destroyed before it manages to update the Neutron database. The reconciliation happens every 60 seconds. We counter this by gracefully deorbiting the Kluster waiting for the updates either by the RouteController or the RouteNanny.

Unfortunately, during normal operations by scaling node pools or terminating nodes updates to the routes do get missed as well. In that case the state in Neutron is correct, while the state on the ASR device still shows the deleted route. This should be fixable by triggering or manually running a re-sync. Unfortunately, that does not work.

The re-sync mechanism between Neutron and ASR is not perfect. There is currently the problem that routes that have been removed in Neutron will not be removed during the sync. It only considers additional routes that have been added.

The only way to recover this situation is to manually `delete` and then `sync` the router using the `asr1k_utils`. Additional node conditions that check for this problem will facilitate alerting and manual intervention.

These duplicate routes are fatal because the IPAM module in the CNI plugin recycles PodCIDRs immediately. A new node will receive the PodCIDR of a previously existing node. The old node's routes are still configured on the ASR device and take precedence. That leaves the new node broken. For example, the state of the ASR routes after multiple node deletions:

```
		198.19.1.0/24 -> 10.180.0.3
		198.19.1.0/24 -> 10.180.0.4
		198.19.1.0/24 -> 10.180.0.5
```

Currently correct is the last route, pointing to the latest instance. In effect is the first route pointing to 10.180.0.3 which doesn't exist anymore.

###  Symptoms

  * See (2). Sporadic Pod/Service/LB Communication Problems
  * Only the first cluster in each project works

### Workarounds:

   * None Known

   * There's no known way to trigger a router `asr1k_utils delete` + `asr1k_utils sync` via OpenStack without actually deleting the whole Neutron router construct. If a side-channel could somehow be leveraged it would be possible to recover automatically.

 ## 5. ASR Missing Routes

Due to various effects it is possible that the ASR agents miss the event to add an additional route when a new node is created.

On specifically fatal effect is the failover between `ACTIVE` and `STANDBY` router. It seems to be a rather common defect (potentially even intended) that only the `ACTIVE` receives sync events. Upon failover routes are missing or reflect the state of a previous cluster.

 ### Symptoms:

  * See (2)

 ### Workarounds:

  * **Manual Sync**
  * **Trigger Sync automatically**
    There's no direct interface to trigger a sync via OpenStack API. It can be forced indirectly by an action that triggers a sync: Attaching/Detaching a FIP/Interface, Adding/Removing a Route

 ## 6. Neutron Static Route Limit
 
 There's a static limit of 31 routes in Neutron.

 In projects with frequent Kluster create and deletes, the route limit can be
 exceeded due duplicate routes as described in (4).

 ### Symptoms:
  
  * See (2). Pod/Service/LB communication problems
  * 409 Conflict Errors in RouteController and RouteNanny
  * Klusters have a max size of 31 Nodes

 ### Workarounds:

  * None. Neutron needs to be reconfigured
  * Clean up duplicate routes

### DIFF
--- a/pkg/client/openstack/kluster/client.go
+++ b/pkg/client/openstack/kluster/client.go
@@ -5,7 +5,10 @@ import (
 	"time"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/secgroups"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	securitygroups "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
 	"github.com/gophercloud/gophercloud/pagination"
 	"k8s.io/client-go/tools/cache"
 
@@ -18,6 +21,8 @@ type KlusterClient interface {
 	CreateNode(*models.NodePool, []byte) (string, error)
 	DeleteNode(string) error
 	ListNodes(*models.NodePool) ([]Node, error)
+	SetSecurityGroup(nodeID string) error
+	EnsureKubernikusRuleInSecurityGroup() (bool, error)
 }
 
 type klusterClient struct {
@@ -115,6 +120,95 @@ func (c *klusterClient) ListNodes(pool *models.NodePool) (nodes []Node, err erro
 	}
 
 	return nodes, nil
+}
+
+func (c *klusterClient) SetSecurityGroup(nodeID string) (err error) {
+	return secgroups.AddServer(c.ComputeClient, nodeID, c.Kluster.Spec.Openstack.SecurityGroupName).ExtractErr()
+}
+
+func (c *klusterClient) EnsureKubernikusRuleInSecurityGroup() (created bool, err error) {
+	page, err := securitygroups.List(c.NetworkClient, securitygroups.ListOpts{Name: c.Kluster.Spec.Openstack.SecurityGroupName}).AllPages()
+	if err != nil {
+		return false, fmt.Errorf("SecurityGroup %v not found: %s", c.Kluster.Spec.Openstack.SecurityGroupName, err)
+	}
+
+	groups, err := securitygroups.ExtractGroups(page)
+	if err != nil {
+		return false, err
+	}
+
+	if len(groups) != 1 {
+		return false, fmt.Errorf("More than one SecurityGroup with name %v found", c.Kluster.Spec.Openstack.SecurityGroupName)
+	}
+
+	udp := false
+	tcp := false
+	icmp := false
+	for _, rule := range groups[0].Rules {
+		if rule.Direction != string(rules.DirIngress) {
+			continue
+		}
+
+		if rule.EtherType != string(rules.EtherType4) {
+			continue
+		}
+
+		if rule.RemoteIPPrefix != c.Kluster.Spec.ClusterCIDR {
+			continue
+		}
+
+		if rule.Protocol == string(rules.ProtocolICMP) {
+			icmp = true
+			continue
+		}
+
+		if rule.Protocol == string(rules.ProtocolUDP) {
+			udp = true
+			continue
+		}
+
+		if rule.Protocol == string(rules.ProtocolTCP) {
+			tcp = true
+			continue
+		}
+
+		if icmp && udp && tcp {
+			break
+		}
+	}
+
+	opts := rules.CreateOpts{
+		Direction:      rules.DirIngress,
+		EtherType:      rules.EtherType4,
+		SecGroupID:     groups[0].ID,
+		RemoteIPPrefix: c.Kluster.Spec.ClusterCIDR,
+	}
+
+	if !udp {
+		opts.Protocol = rules.ProtocolUDP
+		_, err := rules.Create(c.NetworkClient, opts).Extract()
+		if err != nil {
+			return false, err
+		}
+	}
+
+	if !tcp {
+		opts.Protocol = rules.ProtocolTCP
+		_, err := rules.Create(c.NetworkClient, opts).Extract()
+		if err != nil {
+			return false, err
+		}
+	}
+
+	if !icmp {
+		opts.Protocol = rules.ProtocolICMP
+		_, err := rules.Create(c.NetworkClient, opts).Extract()
+		if err != nil {
+			return false, err
+		}
+	}
+
+	return !udp || !tcp || !icmp, nil
 }
 
 func ExtractServers(r pagination.Page) ([]Node, error) {
--- a/pkg/controller/flight/controller.go
+++ b/pkg/controller/flight/controller.go
@@ -1,0 +1,71 @@
+package flight
+
+import (
+	"github.com/go-kit/kit/log"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
+	"github.com/sapcc/kubernikus/pkg/controller/base"
+	"github.com/sapcc/kubernikus/pkg/controller/config"
+)
+
+// =================================================================================
+//   FlightControl
+// =================================================================================
+//
+// This controller takes care about Kluster health. It looks for obvious
+// problems and tries to repair them.
+//
+// Currently implemented are the following helpers. See docs/controllers.md for more
+// in depth explanation why these are required.
+//
+//
+// Delete Incompletely Spawned Instances:
+//
+// It deletes Nodes that didn't manage to register within 10m after
+// inital creation. This is a workaround for DHCP/DVS (latency) issues.  In effect
+// it will delete the incompletely spawned node and launch control will ramp it
+// back up.
+//
+// Ensure Pod-to-Pod Communication via Security Group Rules:
+//
+// It ensures tcp/udp/icmp rules exist in the security group defined during
+// kluster creation. The rules explicitly allow all pod-to-pod
+// communication. This is a workaround for Neutron missing the
+// side-channel security group events.
+//
+//
+// Ensure Nodes belong to the security group:
+//
+// It ensures each Nodes is member of the security group defined in
+// the kluster spec. This ensures missing security groups due to whatever
+// reason are again added to the node.
+
+type FlightController struct {
+	Factory FlightReconcilerFactory
+	Logger  log.Logger
+}
+
+func NewController(threadiness int, factories config.Factories, clients config.Clients, recorder record.EventRecorder, logger log.Logger) base.Controller {
+
+	logger = log.With(logger, "controller", "flight")
+	factory := NewFlightReconcilerFactory(factories.Openstack, factories.NodesObservatory.NodeInformer(), recorder, logger)
+
+	var controller base.Reconciler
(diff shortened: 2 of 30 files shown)


BugRoger:
@fwiesel @abattye Do you have additional insight? Maybe something we missed or miss-interpreted. Tips and tricks?

abattye:
I think the route issues are pretty much as you describe. We have an update from Cisco that might improve the route situation, I haven't applied or teste4d it yet because of the time constraints of the new driver. If this is critical we can try it out in a tame region to see if it improves things. I think there is some locking situation at the root of this.

SchwarzM:
screen shot 2018-03-12 at 09 21 17

you're good but not 2mil. lines good, so i guess there is still vendor crap

pkg/controller/flight/factory.go
```diff
@@ -28,28 +32,38 @@ func (f *flightReconcilerFactory) FlightReconciler(kluster *v1.Kluster) (FlightR
...
 	var reconciler FlightReconciler
-	reconciler = &flightReconciler{kluster, nodes, client, f.Logger}
+	reconciler = &flightReconciler{kluster, instances, nodes, client, f.Logger}
 	reconciler = &LoggingFlightReconciler{reconciler, f.Logger}
 	return reconciler, nil
 }
 
-func (d *flightReconcilerFactory) getNodes(kluster *v1.Kluster, client openstack_kluster.KlusterClient) ([]Node, error) {
-	nodes := []Node{}
+func (d *flightReconcilerFactory) getInstances(kluster *v1.Kluster, client openstack_kluster.KlusterClient) ([]Instance, error) {
```
SchwarzM: maybe we should put the getInstances for whole cluster in the client as well. and not just for single pool

pkg/controller/flight/reconciler.go
```diff
@@ -39,18 +42,70 @@ func (f *flightReconciler) AddMissingSecurityGroup() []string {
...
+		ids = append(ids, instance.GetID())
 	}
 	return ids
 }
 
 func (f *flightReconciler) DeleteIncompletelySpawnedInstances() []string {
-	return []string{}
+	deletedInstanceIDs := []string{}
+	timedOutInstances := f.getTimedOutInstances()
+	unregisteredInstances := f.getUnregisterdInstances()
```
SchwarzM: unregistered ? or unregisterd ?

pkg/controller/flight/reconciler.go
```diff
@@ -39,18 +42,70 @@ func (f *flightReconciler) AddMissingSecurityGroup() []string {
...
+	timedOut := []Instance{}
+	for _, instance := range f.Instances {
+		if instance.GetCreated().Before(time.Now().Add(-INSTANCE_SPAWNING_TIMEOUT)) {
+			timedOut = append(timedOut, instance)
+		}
+	}
+	return timedOut
+}
+
+func (f *flightReconciler) getUnregisterdInstances() []Instance {
```
SchwarzM: see above

BugRoger:
Cleaned up the `vendor` directory again...

pkg/controller/flight/reconciler.go
```diff
@@ -39,18 +42,70 @@ func (f *flightReconciler) AddMissingSecurityGroup() []string {
...
+		found := false
+		for _, timedOut := range timedOutInstances {
+			if unregistered.GetName() == timedOut.GetName() {
+				found = true
+				break
+			}
+		}
+
+		if found {
+			if err := f.Client.DeleteNode(unregistered.GetID()); err != nil {
```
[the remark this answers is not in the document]
BugRoger: This could in deed happen when an instance is stuck in deleting. I'm unsure we should be handling that. It sounds more like a job for a Nova nanny.

If an error occurs this will try to delete it again in the next reconciliation loop. Also, I tried to write funcs to be unaffected by errors and not break the global loop. Hence this only logs the error and `continue`s. It should at most result in 1 log line every 5 minutes, if an instance is stuck or a concurrent deletion happens.

pkg/controller/flight/reconciler.go
BugRoger: fixed

pkg/controller/flight/reconciler.go
BugRoger: fixed

BugRoger:
Rebases, squashed, manually tested. This is ready for review

pkg/controller/flight/controller.go
```diff
@@ -0,0 +1,250 @@
...
+import (
+	"github.com/go-kit/kit/log"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
+	"github.com/sapcc/kubernikus/pkg/controller/base"
+	"github.com/sapcc/kubernikus/pkg/controller/config"
+)
+
+// Flight Control
```
databus23: Awesome docs, but should this maybe also be copied to our markdown docs?

pkg/client/openstack/kluster/client.go
```diff
@@ -117,6 +122,100 @@ func (c *klusterClient) ListNodes(pool *models.NodePool) (nodes []Node, err erro
 	return nodes, nil
 }
 
+func (c *klusterClient) SetSecurityGroup(nodeID string) (err error) {
+	err = secgroups.AddServer(c.ComputeClient, nodeID, c.Kluster.Spec.Openstack.SecurityGroupName).ExtractErr()
+	// ?????
+	if err.Error() == "EOF" {
```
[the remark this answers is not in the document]
BugRoger: Yes, nice. Fixed it with an update.

pkg/client/openstack/kluster/client.go
```diff
@@ -117,6 +122,100 @@ func (c *klusterClient) ListNodes(pool *models.NodePool) (nodes []Node, err erro
...
+		return nil
+	}
+	return err
+}
+
+func (c *klusterClient) EnsureKubernikusRuleInSecurityGroup() (created bool, err error) {
+	page, err := securitygroups.List(c.NetworkClient, securitygroups.ListOpts{Name: c.Kluster.Spec.Openstack.SecurityGroupName}).AllPages()
+	if err != nil {
+		return false, fmt.Errorf("SecurityGroup %v not found", c.Kluster.Spec.Openstack.SecurityGroupName)
```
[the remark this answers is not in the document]
BugRoger: agreed

Issue #164, opened 2018-01-12
Create Private Registry How-To Guide


BugRoger:
Create a guide that outlines how to deploy a private registry on Kubernikus (in CCloud). 

Featuring:

1. Deployment Spec
2. Swift Backend via https://github.com/docker/distribution/pull/2441
3. Expose service via Load Balancer
4. Attach Floating-IP
5. DNS Name
6. SSL Certificate

jonathanschwarze:
I tried different approaches:

**#1 Private Registry Add-on by Kubernetes** (https://github.com/kubernetes/kubernetes/tree/release-1.7/cluster/addons/registry)

There was an addon for private docker registries available. Unfortunately this doesn’t work with Kubernetes > 1.7. The idea was to run a registry in a pod on a node and proxies on all nodes exposing a port to this node via hostPort. Docker sees the registry as localhost therefore no issues with the docker insecure registry policy. 

But it doesn’t work anymore, I ran into issues with the proxies like this:
https://github.com/kubernetes/kubernetes/issues/43986

I couldn’t figure out why they decided to bury this feature. Just some old related issues 

https://github.com/kubernetes/kubernetes/issues/1319
https://github.com/kubernetes/kubernetes/issues/19033

**#2 Helm Docker Registry** (https://github.com/kubernetes/charts/tree/master/stable/docker-registry)

The chart doesn’t work too. 
https://github.com/kubernetes/charts/issues/2991

**#3 Manual installing registry on node**

Another solutions is deploying a registry on a node manually. SSH, tls certs creation, dns etc. Not sure if we should propose this since it could lead to other issues. 

https://docs.docker.com/registry/deploying/#run-an-externally-accessible-registry

PR #248, opened 2018-03-01
Private registry guide 

jonathanschwarze:
https://github.com/sapcc/kubernikus/issues/164


docs/guide/common_addons.md
```diff
@@ -71,4 +71,155 @@ dashboard accessible from the outside via `https` on port `443`.
...
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: registry-storage
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
```
SchwarzM: 1g seems too small.

docs/guide/common_addons.md
```diff
@@ -71,4 +71,155 @@ dashboard accessible from the outside via `https` on port `443`.
...
+        ports:
+        - name: registry
+          containerPort: 80
+          hostPort: 5000
+```
+
+### Access registry from outside
+Through a ssh tunnel you can push or pull images from your cluster registry. At first, export your local ip: 
+```
+export local_ip=$(ipconfig getifaddr en0)
```
SchwarzM: is this working? i thought a tag named 10.46.0.1:5000 in docker is not the same as localhost:5000 is not the same as 127.0.0.1:5000

docs/guide/common_addons.md
jonathanschwarze: True. I'd change it to 5g + add to the description that the value should be adjusted regarding the desired size and available quota.

docs/guide/common_addons.md
jonathanschwarze: The tunnel works fine whereas push/pull via. port-forwarding on Mac doesn't work as expected. Other user are facing the same issue. It seems like forwarded port is not being used by docker for Mac. `Error response from daemon: Get http://localhost:5000/v2/: dial tcp [::1]:5000: getsockopt: connection refused`. The annoying aspect by using the tunnel is to change the local_ip in "insecure repos" on the local machine after getting a new one.

PR #249, opened 2018-03-01
Use the config-drive option to deliver ignition json to the node

SchwarzM:
We are currently dependent on aligning all the stars to get a node to boot-up.
While it works ( as it does now ) but we can alleviate the issues with DHCP and DVS by 
using the config-drive.

OpenStack puts everything it would serve via http into an iso image and attaches it to the vm.
Ignition then recognises the clouddrive and uses it to access it's configuration.

Pro: 
- everytime we have the node configured correctly including passwords
- we do not have to "adapt" the image to ensure a bigger ignition timeout 

Contra:
- we shadow some problems as we will still need to reach the network and dhcp as they are our dns servers as well.

### DIFF
--- a/pkg/client/openstack/client.go
+++ b/pkg/client/openstack/client.go
@@ -613,6 +613,7 @@ func (c *client) CreateNode(kluster *kubernikus_v1.Kluster, pool *models.NodePoo
 
 	name = SimpleNameGenerator.GenerateName(fmt.Sprintf("%v-%v-", kluster.Spec.Name, pool.Name))
 
+	configDrive := true
 	server, err := compute.Create(client, servers.CreateOpts{
 		Name:           name,
 		FlavorName:     pool.Flavor,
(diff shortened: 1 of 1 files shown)


databus23:
If this works I'm all for it. For the remaining network calls we can then tweak the timeouts, right?

SchwarzM:
we will have to make sure that all of our own units:

```Require=network-online.target```
```After=network-online.target```

but it is a quick change on our side and not an image build.

databus23:
This won't work for the baremetal case though, right?

SchwarzM:
uh, that might be true as it highly depends on how the driver is implemented.
most of the bmc's have some kind of usb / emulation / cdrom attachement but i don't know if it is implemented in our drivers ( as we use several across different bm implementations )

BugRoger:
I'm not concerned about bare-metal. As it seems we need to do a special case for passing in multiple networks anyway here.

Issue #190, opened 2018-01-22
missing nodePool status


auhlig:
[
![poolstatus](https://user-images.githubusercontent.com/14873188/35231774-f34d3b4c-ff99-11e7-8aae-4ccfe56d91af.png)
](url)

While showing the status for the 1st nodePool works, it missing for the 2nd+.
It's not Elektra's fault, since the response only contains the status for the 1st nodePool.
NodePool status is not maintained correctly in the kluster resource in the first place, though I can see launctl loops over all nodePools:
```
ts=2018-01-22T16:27:08.222723259Z caller=github.com/sapcc/kubernikus/pkg/controller/base/logging.go:20 controller=launch msg="reconciled kluster" kluster=e2e project=301efa90f09f44f3b0df1c5a2749aac6 requeue=true took=974.901042ms v=1
ts=2018-01-22T16:27:13.582314364Z caller=github.com/sapcc/kubernikus/pkg/controller/launch/logging.go:16 controller=launch kluster=e2e project=301efa90f09f44f3b0df1c5a2749aac6 pool=small msg="read status" running=2 starting=0 stopping=0 needed=0 unneeded=0 took=359.364944ms v=1
ts=2018-01-22T16:27:13.668659217Z caller=github.com/sapcc/kubernikus/pkg/controller/launch/logging.go:33 controller=launch kluster=e2e project=301efa90f09f44f3b0df1c5a2749aac6 pool=small msg="wrote status" running=2 starting=0 stopping=0 needed=0 unneeded=0 took=86.267412ms v=1
ts=2018-01-22T16:27:13.881951575Z caller=github.com/sapcc/kubernikus/pkg/controller/launch/logging.go:16 controller=launch kluster=e2e project=301efa90f09f44f3b0df1c5a2749aac6 pool=prometheus msg="read status" running=1 starting=0 stopping=0 needed=0 unneeded=0 took=213.209527ms v=1
ts=2018-01-22T16:27:13.967668525Z caller=github.com/sapcc/kubernikus/pkg/controller/launch/logging.go:33 controller=launch kluster=e2e project=301efa90f09f44f3b0df1c5a2749aac6 pool=prometheus msg="wrote status" running=1 starting=0 stopping=0 needed=0 unneeded=0 took=85.64902ms v=1
[...]

PR #254, opened 2018-03-05
Fix missing nodepool status

SchwarzM:
this should address #190 

We are currently iterating over all pools in the status but when new pools appear in the spec 
they never get added / updated thus resulting in just the first pool in the status.

### DIFF
--- a/pkg/controller/launch/controller.go
+++ b/pkg/controller/launch/controller.go
@@ -108,7 +108,7 @@ func (lr *LaunchReconciler) terminatePools(kluster *v1.Kluster) (requeue bool, e
 }
 
 func (lr *LaunchReconciler) terminatePool(kluster *v1.Kluster, pool *models.NodePool) (status *PoolStatus, requeue bool, err error) {
-	pm := lr.newPoolManager(kluster, pool)
+	pm := lr.newPoolManager(kluster, pool, lr.klusterInformer.Lister())
 	status, err = pm.GetStatus()
 	if err != nil {
 		return
@@ -126,7 +126,7 @@ func (lr *LaunchReconciler) terminatePool(kluster *v1.Kluster, pool *models.Node
 }
 
 func (lr *LaunchReconciler) reconcilePool(kluster *v1.Kluster, pool *models.NodePool) (status *PoolStatus, requeue bool, err error) {
-	pm := lr.newPoolManager(kluster, pool)
+	pm := lr.newPoolManager(kluster, pool, lr.klusterInformer.Lister())
 	status, err = pm.GetStatus()
 	if err != nil {
 		return
--- a/pkg/controller/launch/controller.go
+++ b/pkg/controller/launch/controller.go
@@ -108,7 +108,7 @@ func (lr *LaunchReconciler) terminatePools(kluster *v1.Kluster) (requeue bool, e
 }
 
 func (lr *LaunchReconciler) terminatePool(kluster *v1.Kluster, pool *models.NodePool) (status *PoolStatus, requeue bool, err error) {
-	pm := lr.newPoolManager(kluster, pool, lr.klusterInformer.Lister())
+	pm := lr.newPoolManager(kluster, pool)
 	status, err = pm.GetStatus()
 	if err != nil {
 		return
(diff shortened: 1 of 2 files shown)


pkg/controller/launch/pool_manager.go
```diff
@@ -113,13 +135,31 @@ func (cpm *ConcretePoolManager) SetStatus(status *PoolStatus) error {
...
+	// Add new pools in the spec to the status
+	for _, specPool := range copy.Spec.NodePools {
+		// Find the pool
+		if npi, ok := nodePoolInfoGet(copy.Status.NodePools, specPool.Name); ok {
+			// is there a need to update?
+			if copy.Status.NodePools[npi] == newInfo {
 				return nil
 			}
+			copy.Status.NodePools[npi] = newInfo
+			_, err = cpm.Clients.Kubernikus.Kubernikus().Klusters(copy.Namespace).Update(copy)
```
databus23: Triggering an update for every change pool seems excessive. I would just set update an `updated` Boolean flag to true and to one update after both ranges.

pkg/controller/launch/pool_manager.go
```diff
@@ -113,13 +135,31 @@ func (cpm *ConcretePoolManager) SetStatus(status *PoolStatus) error {
...
+			return err
```
databus23: returning here would skip the rest of the loop, no? This would be solved by doing only one update in the end and returning that error.

SchwarzM: yes you are right i've leaned too much on the old code and have to revisit the returns.

pkg/controller/launch/pool_manager.go
```diff
@@ -113,13 +135,31 @@ func (cpm *ConcretePoolManager) SetStatus(status *PoolStatus) error {
...
+	}
 
-			copy.Status.NodePools[i] = newInfo
+	// Delete pools from the status that are not in spec
+	for i, infoPool := range copy.Status.NodePools {
+		// find the pool
+		if _, ok := nodePoolSpecGet(copy.Spec.NodePools, infoPool.Name); !ok {
+			// not found in the spec anymore so delete it
+			copy.Status.NodePools = removeNodePool(copy.Status.NodePools, i)
 			_, err = cpm.Clients.Kubernikus.Kubernikus().Klusters(copy.Namespace).Update(copy)
```
BugRoger: We have a helper for updating a kluster. It would require a bit of refactoring though. Maybe as a second step... https://github.com/sapcc/kubernikus/blob/master/pkg/util/kluster.go#L16-L18

SchwarzM:
I think this is ready for another round. thanks for the pointers @databus23.

pkg/controller/launch/controller.go
```diff
@@ -126,7 +126,7 @@ func (lr *LaunchReconciler) terminatePool(kluster *v1.Kluster, pool *models.Node
 }
 
 func (lr *LaunchReconciler) reconcilePool(kluster *v1.Kluster, pool *models.NodePool) (status *PoolStatus, requeue bool, err error) {
-	pm := lr.newPoolManager(kluster, pool)
+	pm := lr.newPoolManager(kluster, pool, lr.klusterInformer.Lister())
```
databus23: See above.

pkg/controller/launch/pool_manager.go
```diff
@@ -109,23 +133,46 @@ func (cpm *ConcretePoolManager) SetStatus(status *PoolStatus) error {
...
+	for _, specPool := range copy.Spec.NodePools {
+		// Find the pool
+		if npi, ok := nodePoolInfoGet(copy.Status.NodePools, specPool.Name); ok {
+			// is there a need to update?
+			if copy.Status.NodePools[npi] == newInfo {
+				continue
 			}
+			copy.Status.NodePools[npi] = newInfo
+			updated = true
+			// not found so add it
```
databus23: move this comment in the `else` block

pkg/controller/launch/pool_manager.go
```diff
@@ -109,23 +133,46 @@ func (cpm *ConcretePoolManager) SetStatus(status *PoolStatus) error {
...
-	for i, curInfo := range copy.Status.NodePools {
-		if curInfo.Name == newInfo.Name {
-			if curInfo == newInfo {
-				return nil
+	updated := false
+	// Add new pools in the spec to the status
+	for _, specPool := range copy.Spec.NodePools {
+		// Find the pool
+		if npi, ok := nodePoolInfoGet(copy.Status.NodePools, specPool.Name); ok {
+			// is there a need to update?
```
databus23: I like it better if the comment describes the `if` case: e.g. "Skip pool if nothing changed"

PR #255, opened 2018-03-06
Custom handling for 403 errors from compute

databus23:
Now that https://github.com/gophercloud/gophercloud/pull/714 is merged upstream we can handle those pesky quota errors more gracefully.


BugRoger:
Nice work!

PR #257, opened 2018-03-06
fixes nodepool update overwriting securitygroup

BugRoger:
Fixes the `u''` problem in the e2e tests. And others...


BugRoger:
@SchwarzM I think, what happened is that we never set a security group when creating a kluster. In `groundctl` there's a sanity check that fills in `default` when nothing is set.

Now, when we update the kluster to add a another NodePool, we also do not put a SecurityGroup into the spec. That deletes the `default` or even another group that has properly set before.

PR #259, opened 2018-03-07
Introduce availabilityZone as OpenStack Metadata auxiliary call

SchwarzM:
- Implemented OpenstackMetadata call to fetch available availabilityzones
- Prepared nodepool to receive availabilityzone attribute


BugRoger:
Something is wrong with the vendoring. There's an update to the vendor folder without changes to glide. Also, oddly specific updates only.

SchwarzM:
that is because we use glide vendor cleanup. it removes subpackages that are not referenced. i have now referenced a new subpackage thus it just "appears".

BugRoger:
Ah, ok. nvm me. carry on... ;)

databus23:
rebases, squashed and reopened...

pkg/client/openstack/project/client.go
```diff
@@ -214,3 +221,23 @@ func (c *projectClient) getFlavors() ([]models.Flavor, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	zones, err := availabilityzones.ExtractAvailabilityZones(allPages)
+	if err != nil {
+		return nil, err
+	}
+	for _, zone := range zones {
+		// only report on available zones
+		if zone.ZoneState.Available {
```
databus23: I'm wondering why only report an available availability zones? The swagger type has a boolean flag to indicate if an AZ is available or not.

BugRoger: Removed both the filter and the boolean from the swagger spec. My rational is that the AVZ will only ever be temporarily unavailable.

I cross-checked that we don't have workaround AVZ that are permanently unavailable. Looks good. So far.

BugRoger:
Tested this locally. It is good to go as a first step to make selection of AVZ possible.

PR #260, opened 2018-03-07
Openstack Client Refactoring

BugRoger:
This is a refactoring and unification of the OpenStack client. 

Unfortunately, this was quite convoluted. Detangling it bit by bit would have resulted in many PRs. This got quite massive now.

The highlights are:
- Shared facility for creating and caching `gophercloud.ProviderClient` and `gophercloud.ServiceClient`
- Seperation into `admin`, `project` and `kluster` clients instead of a single client that only partly works depending on the authentication context
- Clearer Interfaces. Prepared for FakeImplementations for easier testing
- Each OpenStack call is logged
- SharedFactory for creating and caching clients
- `ListNodes` features a ExpiringCache
- Removal of duplicate `GetRouter` functions and models
- GroundControl now uses the `GetMetadata` call for Discovery 
- Cleanup
- Easy to extend with metrics (not yet done)


BugRoger:
Rebased on master

PR #261, opened 2018-03-08
Add quotes to KS_PASSWORD to allow spaces

SchwarzM:
I've got a space in the password which breakes hivemind.


databus23:
Who puts spaces in their passwords???

BugRoger:
> Who puts spaces in their passwords???
Mnemonic passphrase would be my guess ;)

Only the real hardcore nerds are tough enough to use `"` or `'` maybe even a emoji or Chinese character thrown in.

SchwarzM:
innocent

PR #262, opened 2018-03-12
Remove host field from healthz check

SchwarzM:
The healthz check is not issued from within the netns of the pod so 127.0.0.1 seems only to work for host networked containers.


databus23:
Emergency CAP approval granted

PR #267, opened 2018-03-28
Kubernetes v1.10.0

BugRoger:
This PR adds Kubernetes v1.10.0 as default Kubernetes version.

* Updated etcd to recommended version
* Added `ClusterRoleBinding`-Seed to allow the Apiserver to talk to the Kubelets (now required)
* Fixed Coreos-metadata by removing it. It bugs out because we don't set a private-key via Nova. Doesn't work at all. Replaced it with some systemd-foo
* Custom Hyperkube build script

Kubernetes v1.10.0 has a bug crashes the Kubelet on startup. It tries to find it's node-name via the cloud-provider. Someone decided that it's cool to use the `hostname` instead of the `name` from the metadata service. In our setup that includes a domain `.novalocal`. With that added to the node name it can't be found in Nova anymore. In addition the cloud-provider is authorative, e.g. it can't be overwritten. See here: https://github.com/kubernetes/kubernetes/issues/61774

I cherry-picked the fix and build our own image. This gives the added benefit that we don't have to wait for CoreOS to releasing their built for Tectonic. Additionally, we can omit Calico and save ~300MB image size.


BugRoger:
I had to reenable the RouteGC. The cloud-provider doesn't clean up orphaned routes. 🤔

databus23:
🤪 For reference this is my PR that was supposed to fix orphaned routes

BugRoger:
> 🤪 For reference this is my PR that was supposed to fix orphaned routes: kubernetes/kubernetes#56258

Hm. Need to look into what's happening. Maybe the Deorbiter exits quicker than the reconciliation duration is.

databus23:
> Maybe the Deorbiter exits quicker than the reconciliation duration is.

But in that case the orphaned route should be reconciled by the route controller of the next cluster that is created in the same project. As long as there is no new cluster an orphaned route shouldn‘t be a problem.

BugRoger:
> But in that case the orphaned route should be reconciled by the route controller of the next cluster that is created in the same project. As long as there is no new cluster an orphaned route shouldn‘t be a problem.

Ok, then it doesn't do that.

BugRoger:
This is ready to boogie... @SchwarzM @databus23 

Boogie

BugRoger:
Ah, and btw the RouteController in OpenStack is still broken. So, the RouteGC Nanny is activated still.

databus23:
Hopefully this PR will make it into a later 1.10 release finally fixing the openstack route provider. https://github.com/kubernetes/kubernetes/pull/62729

charts/kube-master/templates/api.yaml
```diff
@@ -102,7 +108,6 @@ spec:
             - --etcd-servers=http://{{ include "etcd.fullname" . }}:2379
             - --secure-port=6443
             - --advertise-address={{ .Values.advertiseAddress }}
-            - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota
```
databus23: This was removed because its the default?

pkg/templates/node_1.10.go
```diff
@@ -0,0 +1,385 @@
...
+        [Unit]
+        Description=Kubelet via Hyperkube ACI
+
+        [Service]
+        Environment="RKT_RUN_ARGS=--uuid-file-save=/var/run/kubelet-pod.uuid \
+          --inherit-env \
+          --dns=host \
+          --net=host \
+          --volume var-lib-cni,kind=host,source=/var/lib/cni \
+          --volume var-log,kind=host,source=/var/log \
```
databus23: The /etc/machine-id mount is missing we added with #268 to the other templates

pkg/controller/routegc/controller.go
```diff
@@ -145,7 +145,6 @@ func (r *RouteGarbageCollector) reconcile(kluster *v1.Kluster, logger log.Logger
 }
 
 func (r *RouteGarbageCollector) klusterAdd(obj interface{}) {
-	//TODO: Don't start routegc watchloop for 1.10+ klusters
```
databus23: Maybe instead of removing change the comment for future reference
> TODO: Don't start routegc for kluster versions that have https://github.com/kubernetes/kubernetes/pull/62729 merged

SchwarzM:
please don't use flannel :(

pkg/templates/node_1.10.go
BugRoger: Added.

pkg/controller/routegc/controller.go
BugRoger: Added the TODO again. With 1.11. I'm an optimist ;)

pkg/templates/node_1.10.go
```diff
@@ -41,6 +41,35 @@ systemd:
...
             [Service]
             Environment="DOCKER_OPTS=--log-opt max-size=5m --log-opt max-file=5 --ip-masq=false --iptables=false --bridge=none"
+    - name: flanneld.service
+      enable: true
+      dropins:
+        - name: 10-ccloud-opts.conf
+          contents: |
+            [Service]
+            ExecStartPre=/usr/bin/systemctl is-active ccloud-metadata-hostname.service
+            ExecStartPre=/usr/bin/bash -c "/usr/bin/systemctl set-environment NODE_NAME=$(hostname -s)"
```
databus23: That hurts a little.

pkg/templates/node_1.10.go
```diff
@@ -41,6 +41,35 @@ systemd:
...
+                                      -kubeconfig-file=/var/lib/kubelet/kubeconfig \
+                                      -kube-api-url={{ .ApiserverURL }}"
+            Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/coreos/flannel-wrapper.uuid \
+                                      --volume var-lib-kubelet,kind=host,source=/var/lib/kubelet,readOnly=true \
+                                      --mount volume=var-lib-kubelet,target=/var/lib/kubelet \
+                                      --volume etc-kubernetes-certs,kind=host,source=/etc/kubernetes/certs,readOnly=true \
+                                      --mount volume=etc-kubernetes-certs,target=/etc/kubernetes/certs \
+                                      --volume etc-kube-flannel,kind=host,source=/etc/kube-flannel,readOnly=true \
+                                      --mount volume=etc-kube-flannel,target=/etc/kube-flannel"
+    - name: flannel-docker-opts.service
```
databus23: Is this an override,  so that flannel does nothing with docker?

BugRoger: Yes, the `flannel-docker-opts.service` is locked in sync with `flanneld.service`. Both need to be able to start. It also wants to start before `docker` so that it can inject it's config. Without `docker` the `kubelet` won't start. If that doesn't start the node never registers itself and `flannel` doesn't find via the API and crashes. It's a nice dead-lock.

What I also learned is that systemd ignores a `restart=always` in a unit file when the service fails for any reason other than `Exec.*` fails with an error. Here it throws a dependency error and stops completely. Otherwise this convolution would maybe accidentally start eventually.

Anyway, we don't want the Docker options, so I'm overwriting it with a noop service. Flannel still crashes until the node registers itself. But now it actually restarts. Once the node is registered it also gets its `podCIDR`, so when Flannel comes up finally it's already set and reused.

BugRoger:
Yes, agreed. It's a bit convoluted. I think though that it's mostly due to the `rkt-flannel-wrapper` stuff, which brings the risk that it changes unexpected with an CoreOS update. And that wrapper pulls in the unwanted docker options side-effects.

We could:
- Write our own wrapper that only does what we want and starts correctly
- Copy-Paste the logic to the wormhole
- Just try and see. Change again when needed

BugRoger:
I'm quiet happy that we came up with this idea. It's a workable solution to get this running in `eu-de-1`. Also a nice test whether we can leverage it to replace the dynamic BGP PodCIDR announcements.

BugRoger:
Added `NODE_NAME` into `/etc/kubernetes/environment`. Squashed and rebased.

PR #270, opened 2018-04-17
Less Password for Everyone

BugRoger:
This PR make it that `kubernikusctl` stores entered passwords in the OS' keychain service.

PasswordPolicy

No more typing the same password 100s of times!

### DIFF
--- a/pkg/cmd/kubernikusctl/common/openstack.go
+++ b/pkg/cmd/kubernikusctl/common/openstack.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	keyring "github.com/zalando/go-keyring"
 )
 
 type OpenstackClient struct {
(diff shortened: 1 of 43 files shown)


BugRoger:
> On other platforms, especially linux this might throw errors, when there is no keyring subsystem available.

For that reason, the error here is silently swept under the rug. https://github.com/sapcc/kubernikus/pull/270/files#diff-ea17b6dc668c598ccc9960576b46a659R130

pkg/cmd/kubernikusctl/common/openstack.go
```diff
@@ -99,7 +100,20 @@ func (o *OpenstackClient) Setup() error {
...
-		o.Password = os.Getenv("OS_PASSWORD")
+		if os.Getenv("OS_PASSWORD") != "" {
+			o.Password = os.Getenv("OS_PASSWORD")
+		} else {
+			username := os.Getenv("USER")
+			if o.Username != "" {
+				username = o.Username
+			}
+
+			password, err := keyring.Get("kubernikus", username)
```
databus23: But wouldn't this maybe fail when there is no keyring manager available?

MichaelSp:
this just bite me
```bash
> kubernikusctl auth refresh
An error occurred: exec: "dbus-launch": executable file not found in $PATH
```

running in a docker container.

PR #272, opened 2018-04-20
Stop! Hammer time.

databus23:
This PR adds a `hammertime` controller.

Here is a short video explaining what it does:

[Stop Hammertime](https://www.youtube.com/watch?v=H0_ICNMbdzw)

It uses the nodeobservatory to monitor the nodes of all klusters.
If for a kluster no node has posted an update within 20 seconds it instructs the cluster to have a little hammer time.
In hammer time the controller manager is shutdown to avoid having the service controller remove all nodes from loadbalancer.

We do this to keep the payload running and accessible should a network problem between apiserver and kubelets prevent nodes from posting their status. This is especially usefully for our cross region setups.

### DIFF
--- a/pkg/cmd/kubernikus/operator.go
+++ b/pkg/cmd/kubernikus/operator.go
@@ -48,7 +48,7 @@ func NewOperatorOptions() *Options {
 	options.KubernikusDomain = "kluster.staging.cloud.sap"
 	options.Namespace = "kubernikus"
 	options.MetricPort = 9091
-	options.Controllers = []string{"groundctl", "launchctl", "deorbiter", "routegc", "flight", "migration"}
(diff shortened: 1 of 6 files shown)


BugRoger:
🤣 Approved! So few pixels in the 80s...

BugRoger:
Y No merged??

databus23:
@BugRoger Still isn't approved yet...

databus23:
reopen. This is still valid.

BugRoger:
Approved x2

databus23:
I added a single metric to track when a cluster has a little hammer time. Merging this now

BugRoger:
Nice. Finally!

Hammer Time

PR #274, opened 2018-04-25
Lower-Case Node Name for Openstack

BugRoger:
Apparently, the cloud-provider in v1.10 became case-sensitive when it comes to node names. It infers the node's name via the hostname provided through the metadata service. A hostname is always lowercase.  Not so much the node names in Nova.

It was possible to create node pools with uppercase letters, which in turn creates nodes like `kluster-Pool-xyz123` their hostname is then `kluster-pool-xyz123`. The Kubelet can't register that node anymore.


BugRoger:
anyone?

PR #290, opened 2018-06-22
Replaces glog with glognomore

### LINKED ISSUE #285: Make glog log to gokit-log
https://github.com/kubermatic/glog-gokit

BugRoger:
We use a trick in Glide to overwritte the glog package with a custom
implementation. The modified shim implements the glog interface but
actually logs to gokit-log.

Fixes #285


BugRoger:
Good idea. I added a key `glog` to each function call in the shim. It also transports the severity that was used to call in glog.

e.g. `glog=Fatal`, `glog=Warning`

Should be possible to `grep -v glog=` now and filter out the remaining glog spam.

databus23:
Nice!

PR #291, opened 2018-06-25
Adds quota checks pre/post e2e test runs

BugRoger:
This commit adds a quota check before the e2e test are started. The test
fail with a concrete error message which quota was exceeded. That
safe-guards against obscure timeouts that happen to to missing quota.

After the e2e test have been completed, there's now also an additional
post-flight quota check. It will fail the test if any resource is still
left over and hasn't been cleaned up properly.

This should make it easier to identify when, where and how the tests are
leaking resources.

In case an existing kluster is reused for e2e tests, the quota checks
are skipped.

### DIFF
--- a/deps/deps.go
+++ b/deps/deps.go
@@ -4,6 +4,8 @@
 package extra_dependencies
 
 import (
+	_ "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/quotasets"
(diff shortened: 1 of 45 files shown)


BugRoger:
paging @auhlig. Trading 🍪 for 👍.

PR #292, opened 2018-06-27
Clean-Up Errored Instances

BugRoger:
This PR adds functionality to the flight controller that deletes
instances in Error state. Recreation works in tandem with launch control
which just ignores errored VMs.


BugRoger:
LGTM

PR #308, opened 2018-08-07
Have the kublet self label the nodes during registration

### LINKED ISSUE #305: Automatically Label GPU Nodes
Implement a mechanism that automatically labels GPU nodes.

We need a label:
`ccloud.sap.com/nvidia-gpu=nvidia-tesla-v100`

Ideally this would pick the correct model of GPU from the node's flavor. Currently the flavor doesn't hold this information. 

For now we need to hard-code flavors starting with `zg*`, like `zg1bcm1.medium` with the value `vidia-tesla-v100`

databus23:
This always adds a node label `ccloud.sap.com/nodepool` and for nodes with flavour zg* it also adds `ccloud.sap.com/nvidia-gpu`.

Fixes #305

Note: This is using the kubelet flag `--node-labels` which is still alpha and potentially will change in the future: https://github.com/kubernetes/community/pull/911.

I still went with this approach as its considerable less effort for now and as I understand it from various GitHub comments, gke is using the same flag for their gpu nodes as of now.

I modified the integration test to ensure the nodepool label is present, so that we catch any changes to the node-labels flag with future kubernetes versions

### DIFF
--- a/pkg/controller/launch/pool_manager.go
+++ b/pkg/controller/launch/pool_manager.go
@@ -190,7 +190,7 @@ func (cpm *ConcretePoolManager) CreateNode() (id string, err error) {
 
 	nodeName := util.SimpleNameGenerator.GenerateName(fmt.Sprintf("%v-%v-", cpm.Kluster.Spec.Name, cpm.Pool.Name))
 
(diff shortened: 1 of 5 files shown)


BugRoger:
Nice! Thanks... 😃

PR #312, opened 2018-08-16
Add per nodepool servergroup with soft-affinity

### LINKED ISSUE #309: Cinder Affinity Workaround
In order to avoid excessive waiting times and potential data-loss all nodes should be co-located on the same building block.

We need to ensure a "Server Group" with an `affinity` policy exists for each project. https://developer.openstack.org/api-ref/compute/#create-server-group

Then use the ID of that server group when creating new nodes. The server call accepts an array `os:scheduler_hints` with a value of `group=a0cf03a5-d921-4877-bb5c-86d26cf818e1`

This feature can be hardcoded for now. Eventually, we can add an `affinity` flag to the pool options.

databus23:
This PR creates an openstack server group for every node pool (named $KLUSTER/$NODEPOOL) with policy `soft-affinity`. This ensures that all instances of a node pool are placed within a single VMware building block.

For single node pool clusters this should fix our volume re-attachment problems. For multiple nodepools users need to be aware that volume attachments are guaranteed only to by fast within a node pool and they should use `nodeSelectors` accordingly.

One open quibble I have with this is that its pretty custom to our openstack deployment (using VMware) and probably should be behind a feature flag. In the kvm case this will place all instances of a node pool on a single hypervisor which is most certainly not wanted.

Fixes #309

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-operator: bin/darwin/kubernikus operator --auth-username=$KS_USERNAME --auth-domain=$KS_USER_DOMAIN_NAME --auth-password="$KS_PASSWORD" --auth-project=$KS_PROJECT_NAME --auth-project-domain=$KS_PROJECT_DOMAIN_NAME --auth-url=$KS_AUTH_URL --namespace=$KS_NAMESPACE --context=$KS_CONTEXT --kubernikus-domain=$KS_DOMAIN --v=5
+operator: env NODEPOOL_AFFINITY=true bin/darwin/kubernikus operator --auth-username=$KS_USERNAME --auth-domain=$KS_USER_DOMAIN_NAME --auth-password="$KS_PASSWORD" --auth-project=$KS_PROJECT_NAME --auth-project-domain=$KS_PROJECT_DOMAIN_NAME --auth-url=$KS_AUTH_URL --namespace=$KS_NAMESPACE --context=$KS_CONTEXT --kubernikus-domain=$KS_DOMAIN --v=5
 api: bin/darwin/apiserver --context=$KS_CONTEXT --namespace=$KS_NAMESPACE --auth-url=$KS_AUTH_URL --v=5
--- a/pkg/controller/launch/controller.go
+++ b/pkg/controller/launch/controller.go
@@ -128,6 +128,9 @@ func (lr *LaunchReconciler) terminatePool(kluster *v1.Kluster, pool *models.Node
 			return
 		}
 	}
+	if err = pm.DeletePool(); err != nil {
+		return
+	}
 
 	err = pm.SetStatus(status)
 	return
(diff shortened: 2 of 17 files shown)


BugRoger:
> For single node pool clusters this should fix our volume re-attachment problems. For multiple nodepools users need to be aware that volume attachments are guaranteed only to by fast within a node pool and they should use nodeSelectors accordingly.

This will be a nice source of confusion. It's definitely an odd quirk. I'm debating whether adding the availability zone to the pool would make it better or worse. One one side it's more obvious that the user needs to care about node-selection. On the other hand, it's even more complex to get right.

pkg/controller/launch/controller.go
```diff
@@ -128,6 +128,9 @@ func (lr *LaunchReconciler) terminatePool(kluster *v1.Kluster, pool *models.Node
 			return
 		}
 	}
+	if err = pm.DeletePool(); err != nil {
```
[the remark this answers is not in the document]
databus23: Yes this was meant to be idempotent

pkg/controller/launch/instrumenting.go
```diff
@@ -126,5 +126,5 @@ func (pm *InstrumentingPoolManager) DeleteNode(id string) (err error) {
 }
 
 func (pm *InstrumentingPoolManager) DeletePool() (err error) {
-	return pm.DeletePool()
+	return pm.PoolManager.DeletePool()
```
BugRoger: Nice catch

Procfile
```diff
@@ -1,2 +1,2 @@
-operator: bin/darwin/kubernikus operator --auth-username=$KS_USERNAME --auth-domain=$KS_USER_DOMAIN_NAME --auth-password="$KS_PASSWORD" --auth-project=$KS_PROJECT_NAME --auth-project-domain=$KS_PROJECT_DOMAIN_NAME --auth-url=$KS_AUTH_URL --namespace=$KS_NAMESPACE --context=$KS_CONTEXT --kubernikus-domain=$KS_DOMAIN --v=5
+operator: env NODEPOOL_AFFINITY=true bin/darwin/kubernikus operator --auth-username=$KS_USERNAME --auth-domain=$KS_USER_DOMAIN_NAME --auth-password="$KS_PASSWORD" --auth-project=$KS_PROJECT_NAME --auth-project-domain=$KS_PROJECT_DOMAIN_NAME --auth-url=$KS_AUTH_URL --namespace=$KS_NAMESPACE --context=$KS_CONTEXT --kubernikus-domain=$KS_DOMAIN --v=5
```
BugRoger: Isn't this also missing from the `kubernikus` chart?

BugRoger: Oh. NVM me. Approved. LGTM. Move along...

databus23:
rebased on master. merging when the tests complete

Issue #304, opened 2018-08-07
Add ExtendedResourceToleration AdmissionController


BugRoger:
We need an additional admission controller in the Apiserver. Background: https://notes.mindprince.in/2017/12/17/dedicated-node-pools-and-ExtendedResourceToleration-admission-controller.html

e.g.
`- --enable-admission-plugins=ExtendedResourceToleration`

BugRoger:
Fixed by #315

PR #315, opened 2018-09-14
Taint gpu nodes with nvidia.com/gpu=present:NoSchedule

dhague:
This follows the pattern of GCP, and prevents non-GPU payloads being scheduled to expensive GPU nodes


dhague:
Fixes #304

pkg/templates/ignition.go
```diff
@@ -95,6 +95,11 @@ func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, pool *models.Node
 		}
 	}
 
+	var nodeTaints []string
+	if pool != nil && strings.HasPrefix(pool.Flavor, "zg") {
```
BugRoger: I'm not a fan of this. But it's not your problem. We need to expose the node constraints into the flavor properties instead of encoding it into the flavor name. We already have precedent of doing it like this too.

pkg/templates/ignition_test.go
```diff
@@ -90,5 +90,6 @@ func TestNodeLabels(t *testing.T) {
 	if assert.NoError(t, err, "Failed to generate node") {
 		//Ensure we rendered the expected template
 		assert.Contains(t, string(data), fmt.Sprintf("--node-labels=ccloud.sap.com/nodepool=%s,gpu=nvidia-tesla-v100", pool.Name))
+		assert.Contains(t, string(data), fmt.Sprintf("--register-with-taints=nvidia.com/gpu=present:NoSchedule"))
```
BugRoger: +1 Bonus points for adding a test 😄

PR #319, opened 2018-09-20
alert on many klusters pending

auhlig:
Currently we trigger an alert if a kluster is stuck in `phase!="Running`, which means we get an alert i f a kluster is stuck in `pending`. In most cases due to a project-specific configuration error. Nothing todo from our side. Also it's not an indication that kubernikus did something wrong.

Proposal: Alert only if more than 20% of the klusters are in pending, as this might indicate a more general problem.

[] Yes
[] No
[] Maybe

@BugRoger @databus23

### DIFF
--- a/charts/kubernikus-system/charts/prometheus/kubernikus.alerts
+++ b/charts/kubernikus-system/charts/prometheus/kubernikus.alerts
@@ -2,7 +2,7 @@ groups:
 - name: kubernikus.alerts
   rules:
   - alert: KubernikusKlusterStuck
-    expr: kubernikus_kluster_status_phase{phase!="Running"} == 1
+    expr: kubernikus_kluster_status_phase{phase!~"Running|Pending"} == 1
     for: 1h
     labels:
       tier: kubernikus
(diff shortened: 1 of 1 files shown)


charts/kubernikus-system/charts/prometheus/kubernikus.alerts
```diff
@@ -2,7 +2,7 @@ groups:
 - name: kubernikus.alerts
   rules:
   - alert: KubernikusKlusterStuck
-    expr: kubernikus_kluster_status_phase{phase!="Running"} == 1
+    expr: kubernikus_kluster_status_phase{phase!~"Running|Pending"} == 1
```
BugRoger: IMO it would be clearer without negation.

```suggestion
    expr: kubernikus_kluster_status_phase{phase=~"Creating|Terminating"} == 1
```

charts/kubernikus-system/charts/prometheus/kubernikus.alerts
```diff
@@ -13,6 +13,18 @@ groups:
       description: Kluster {{ $labels.kluster_id }} is stuck in {{ $labels.phase }} for 1h
       summary: Kluster stuck in phase {{ $labels.phase }}
 
+  - alert: KubernikusManyKlusterStuckInPending
+    expr: round((count(kubernikus_kluster_status_phase{phase="Pending"} == 1)) / count(kubernikus_kluster_info) * 100) >= 20
```
BugRoger: This would rarely fire. Assume we have 50 klusters. We would need to have 10 `pending` clusters until we see the alert.

```suggestion
 - alert: KubernikusKlusterStuckInPending
    expr: kubernikus_kluster_status_phase{phase="Pending"} == 1
    for: 168h
    labels:
      tier: kubernikus
      service: k8s
      severity: info
      context: kluster
    annotations:
      description: Kluster {{ $labels.kluster_id }} is stuck in {{ $labels.phase }} for more than a week. Check the project and delete abandoned klusters.
      summary: Kluster stuck in phase {{ $labels.phase }}
```

PR #321, opened 2018-09-25
detect OOM killed pods via NPD

auhlig:
introduce node-problem-detector (NPD) to grep kernel log for oom killed pods.
will rollout, test and introduce an alert later. meanwhile you'll see the status condition on the node as well as the event
```
$ k describe node <node>
Events:
  Type     Reason         Age   From                                     Message
  ----     ------         ----  ----                                     -------
  Warning  PodOOMKilling  2m    kernel-monitor, <node>  Memory cgroup out of memory: Kill process 25169 (hyperkube) score 2109 or sacrifice child
```
LGTY @databus23?

### DIFF
--- a/charts/kubernikus-system/charts/node-problem-detector/plugins/kernel-monitor.json
+++ b/charts/kubernikus-system/charts/node-problem-detector/plugins/kernel-monitor.json
@@ -1,0 +1,68 @@
+{
+  "plugin": "journald",
+  "pluginConfig": {
+    "source": "kernel"
+  },
+  "logPath": "/var/log/journal",
+  "lookback": "5m",
+  "bufferSize": 10,
+  "source": "kernel-monitor",
+  "conditions": [
+    {
+      "type": "KernelDeadlock",
+      "reason": "KernelHasNoDeadlock",
+      "message": "kernel has no deadlock"
+    },
+    {
+      "type": "PodOOMKilling",
+      "reason": "NoPodOOMKilling",
+      "message": "no pods were killed after being OOM"
+    }
+  ],
+  "rules": [
+    {
+      "type": "temporary",
+      "reason": "OOMKilling",
+      "pattern": "Kill process \\d+ (.+) score \\d+ or sacrifice child\\nKilled process \\d+ (.+) total-vm:\\d+kB, anon-rss:\\d+kB, file-rss:\\d+kB"
+    },
+    {
+      "type": "temporary",
+      "condition": "PodOOMKilling",
+      "reason": "PodOOMKilling",
+      "pattern": "\\w+ cgroup out of memory: Kill process \\d+ (.+) score \\d+ or sacrifice child"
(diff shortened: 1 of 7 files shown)


charts/kubernikus-system/charts/node-problem-detector/plugins/kernel-monitor.json
```diff
@@ -0,0 +1,68 @@
...
+  "bufferSize": 10,
+  "source": "kernel-monitor",
+  "conditions": [
+    {
+      "type": "KernelDeadlock",
+      "reason": "KernelHasNoDeadlock",
+      "message": "kernel has no deadlock"
+    },
+    {
+      "type": "PodOOMKilling",
```
auhlig: new status condition `PodOOMKilling`

charts/kubernikus-system/charts/node-problem-detector/plugins/kernel-monitor.json
```diff
@@ -0,0 +1,68 @@
...
+    {
+      "type": "temporary",
+      "reason": "OOMKilling",
+      "pattern": "Kill process \\d+ (.+) score \\d+ or sacrifice child\\nKilled process \\d+ (.+) total-vm:\\d+kB, anon-rss:\\d+kB, file-rss:\\d+kB"
+    },
+    {
+      "type": "temporary",
+      "condition": "PodOOMKilling",
+      "reason": "PodOOMKilling",
+      "pattern": "\\w+ cgroup out of memory: Kill process \\d+ (.+) score \\d+ or sacrifice child"
```
auhlig: the pattern we're looking for. 
another option would be something like `Memory cgroup stats for /kubepods/burstable/podfab2a136-9a7a-11e8-8f56-46884b7d303f`

charts/kubernikus-system/charts/node-problem-detector/templates/daemonset.yaml
```diff
@@ -0,0 +1,64 @@
...
+spec:
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: node-problem-detector
+    spec:
+      tolerations:
+      - key: "species"
```
BugRoger: This looks like a leftover from the bare-metal deployment. How do I even read this? If the key `species` exists it adds a `NoSchedule` toleration? Confused.

auhlig: Indeed. Removed it. Not sure we even need it in the bare-metal  deploy.

Read like: If any node taint `Exists` that would cause a `NoSchedule` of the pod it is ignored. In other words, the pods tolerate any node taint and get deployed everywhere.

BugRoger: Wat? 🤣

PR #324, opened 2018-10-08
Add fluent-bit dependency and config values

jknipper:
Adds fluent-bit chart dependency and basic configuration values to kubernikus-system.


charts/kubernikus-system/requirements.yaml
```diff
@@ -14,3 +14,6 @@ dependencies:
   - name: prometheus-node-exporter
     repository: https://kubernetes-charts.storage.googleapis.com/
     version: 0.3.0
+  - name: fluent-bit
```
databus23: I would like it better if we vendor forked charts in the `charts/` subdirectory. Its more approachable to further changes and also much easier to take a look what the cart actually contains.

jknipper:
In upstream config template is broken. The following PRs exists regarding this topic:
https://github.com/helm/charts/pull/8166
https://github.com/helm/charts/pull/7878
https://github.com/helm/charts/pull/7694
https://github.com/helm/charts/pull/7413
For now only kubernetes container logs are working. It seems we have systemd journal already in place in /var/log/journal and kernel log is in there as well. We need to adapt input config to log this to Elasticsearch as well. I'm going to move the chart here and change config accordingly.

jknipper:
I added the chart to the repo with fixed config template and systemd logging capabilities.

jknipper:
@databus23 I added official fluent-bit chart dependency, configmap and values for configuration.

jknipper:
> @jknipper before we can merge this we need to add the required values to cc/secrets. Can you create a PR for that?

It's already there.

PR #325, opened 2018-10-15
Adds backup/restore sidecar to etcd

jknipper:
This PR adds backup/restore capabilities to etcd, see https://github.com/gardener/etcd-backup-restore
The backup is stored in swift and gets restored if etcd data is corrupted somehow.


charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -42,33 +76,60 @@ spec:
...
+            - --garbage-collection-period-seconds=60
+          image: eu.gcr.io/gardener-project/gardener/etcdbrctl:0.3.0
+          imagePullPolicy: Always
+          ports:
+            - containerPort: 8080
+              name: server
+              protocol: TCP
+          env:
+            - name: STORAGE_CONTAINER
+              value: etcd-backup-{{ .Values.klustername }}-{{ uuidv4 }}
```
BugRoger: This will get pretty long. Did you check this actually works? 😛 

Also, when does `uuidv4` change? Couldn't it happen that we change something in this Helm chart and we get a new UID, losing the old backup? Not support critical but potentially a source of many garbage backups.

charts/kube-master/charts/etcd/templates/secrets.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: Secret
+metadata:
+  name: {{ include "fullname" . }}
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    release: {{ .Release.Name }}
+type: Opaque
+data:
+  openstack-auth-url: {{ required "missing openstack-auth-url" .Values.openstack.authURL | b64enc }}
+  openstack-username: {{ required "missing openstack-username" .Values.openstack.username | b64enc }}
```
BugRoger: I think, we ultimately end up with a service user here. Does it have permissions to do the swift operations? I'm unsure. It might only have the `network_admin` role. Please check.

charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -42,33 +76,60 @@ spec:
...
             - mountPath: /var/lib/etcd
               name: data
+            - mountPath: /bootstrap
+              name: {{ include "fullname" . }}-bootstrap
           resources:
+        - name: backup
+          command:
+            - etcdbrctl
+            - server
+            - --schedule=*/5 * * * *
```
BugRoger: Any idea what the storage requirements are?

charts/kube-master/charts/etcd/templates/secrets.yaml
```diff
@@ -0,0 +1,15 @@
...
+  openstack-password: {{ required "missing openstack-password" .Values.openstack.password | b64enc }}
+  openstack-domain-name: {{ required "missing openstack-domain-name" .Values.openstack.domainName | b64enc }}
+  openstack-project-id: {{ required "missing openstack-project-id" .Values.openstack.projectID | b64enc }}
```
BugRoger: What project do we end up having configured here? Is it our control-plane project or the customers project?

charts/kube-master/charts/etcd/templates/deployment.yaml
jknipper: This will generate something like "etcd-backup-mykluster-4f1d1d5a-d845-4bfb-940b-be6b444497a1". We need a kluster-unique value here to identify the correct backup set. This might become a problem when upgrading helm chart. Any ideas?

charts/kube-master/charts/etcd/templates/deployment.yaml
jknipper: Hard to tell, this depends on general etcd data amount and utilization/changes in data.

charts/kube-master/charts/etcd/templates/deployment.yaml
BugRoger: `etcd-backup-mykluster-4f1d1d5a-d845-4bfb-940b` is already unique. At least as long as the kluster isn't deleted and recreated with the same name.

charts/kube-master/charts/etcd/templates/secrets.yaml
jknipper: Both, customer backups should be on customer quota.

pkg/client/openstack/admin/client.go
```diff
@@ -251,3 +260,54 @@ func (c *adminClient) getRoleID(roleName string) (string, error) {
...
+	}
+
+	allEndpoints, err := endpoints.ExtractEndpoints(allPages)
+	if err != nil {
+		return err
+	}
+
+	var swiftEndpoint endpoints.Endpoint
+	for _, endpoint := range allEndpoints {
+		if endpoint.Name == "swift" {
```
databus23: I would filter by `type` instead. The name is arbitrarily choosen.
```suggestion
		if endpoint.Type == "objectstore" {
```

pkg/client/openstack/admin/client.go
```diff
@@ -251,3 +260,54 @@ func (c *adminClient) getRoleID(roleName string) (string, error) {
 	return "", fmt.Errorf("Role %s not found", roleName)
 
 }
+
+func (c *adminClient) CreateContainer(containerName, serviceUserName, serviceUserDomainName, serviceUserProjectID string) error {
```
databus23: the `projectID` is not tied to the serviceUser we want to grant access. I fact the user is from another domain even. So I would name it differently.
```suggestion
func (c *adminClient) CreateContainer(projectID string, containerName, serviceUsername, serviceUserDomainName) error {
```

pkg/client/openstack/admin/client.go
```diff
@@ -251,3 +260,54 @@ func (c *adminClient) getRoleID(roleName string) (string, error) {
...
+		return err
+	}
+
+	acl := fmt.Sprintf("%s:%s", serviceUserProjectID, serviceUser.ID)
+	createOpts := containers.CreateOpts{
+		ContainerRead:  acl,
+		ContainerWrite: acl,
+	}
+
+	res := containers.Create(storageClient, containerName, createOpts)
```
databus23: I believe this should be like this instead:
```suggestion
	_, err == containers.Create(storageClient, containerName, createOpts).Extract()
```

pkg/client/openstack/admin/client.go
```diff
@@ -23,25 +26,31 @@ type AdminClient interface {
 	DeleteUser(username, domainName string) error
 	GetKubernikusCatalogEntry() (string, error)
 	GetRegion() (string, error)
+	CreateContainer(containerName, serviceUserName, serviceUserDomainName, serviceUserProjectID string) error
```
databus23: Maybe we should call this method `CreateStorageContainer` to make it clearer which type of container.

pkg/controller/ground.go
```diff
@@ -418,6 +419,15 @@ func (op *GroundControl) createKluster(kluster *v1.Kluster) error {
 		Region:     region,
 	}
 
+	if err := op.Clients.OpenstackAdmin.CreateContainer(
+		fmt.Sprintf(etcd_util.EtcdBackupStorageContainer, kluster.GetName(), kluster.GetUID()),
```
databus23: On closer look maybe we should move this format call into a function in the etcd util package and use the function in both places:
`func DefaultStorageContainer(kluster *v1.Kluster) string`

pkg/util/helm/helm.go
```diff
@@ -111,6 +113,7 @@ func KlusterToHelmValues(kluster *v1.Kluster, openstack *OpenstackOptions, certi
 			Persistence: persistenceValues{
 				AccessMode: accessMode,
 			},
+			StorageContainer: fmt.Sprintf(etcd_util.EtcdBackupStorageContainer, kluster.GetName(), kluster.GetUID()),
```
databus23: See my comment above

pkg/client/openstack/admin/client.go
jknipper: @databus23 It seems that Endpoint.Type is not available, see https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/endpoints/results.go#L41

jknipper:
@databus23 I'm now fetching the endpoint by service id, please take a look.

charts/kube-master/charts/etcd/templates/secrets.yaml
databus23: It doesn't need to we create the swift container using the admin client und give the cluster service user access to only this swift container using swits acls

pkg/util/helm/helm.go
```diff
@@ -38,7 +39,8 @@ type persistenceValues struct {
 }
 
 type etcdValues struct {
-	Persistence persistenceValues `yaml:"persistence,omitempty"`
+	Persistence      persistenceValues `yaml:"persistence,omitempty"`
+	StorageContainer string            `yaml:"storageContainer,omitempty"`
```
databus23: I think we are missing the 5 openstack service user fields here we need for the etcd secret ( see above).

jknipper:
@databus23 I added those missing fields.

pkg/util/helm/helm.go
```diff
@@ -114,6 +115,17 @@ func KlusterToHelmValues(kluster *v1.Kluster, openstack *OpenstackOptions, certi
...
 			},
 			StorageContainer: etcd_util.DefaultStorageContainer(kluster),
+			Openstack: openstackValues{
+				AuthURL:             openstack.AuthURL,
+				Username:            openstack.Username,
+				Password:            openstack.Password,
+				DomainName:          openstack.DomainName,
+				Region:              openstack.Region,
+				ProjectID:           kluster.Spec.Openstack.ProjectID,
+				LbSubnetID:          kluster.Spec.Openstack.LBSubnetID,
```
databus23: Those three values are not needed, I would remove them

jknipper: @databus23 I removed the unneeded values.

jknipper:
@databus23 I added some fixes and deferred deletion of backups, please check.

pkg/client/openstack/admin/client.go
```diff
@@ -251,3 +261,76 @@ func (c *adminClient) getRoleID(roleName string) (string, error) {
...
+		ContainerRead:  acl,
+		ContainerWrite: acl,
+	}
+
+	_, err = containers.Create(storageClient, containerName, createOpts).Extract()
+
+	return err
+}
+
+func (c *adminClient) GetPublicObjectStoreEndpointURL(projectID string) (string, error) {
```
databus23: Can you please make this a private function of the adminClient and remove it from the AdminClient interface for now?

pkg/util/etcd/etcd.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
+)
+
+const (
+	EtcdBackupStorageContainer = "etcd-backup-%s-%s"
+)
+
+func DefaultStorageContainer(kluster *v1.Kluster) string {
+	return fmt.Sprintf(EtcdBackupStorageContainer, kluster.GetName(), kluster.GetUID())
```
databus23: `GetName()` returns the full name of the kluster CRD ($NAME-$PROJECTID). As the storage container is created inside the project I think we can safely remove the project id from the container name and just go with $NAME-$UID
```suggestion
	return fmt.Sprintf(EtcdBackupStorageContainer, kluster.Spec.Name, kluster.GetUID())
```

charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -19,21 +53,17 @@ spec:
         release: {{ .Release.Name }}
     spec:
       volumes:
-        {{- if false }}
```
databus23: bonus points for removing this crap 👍

charts/kube-master/charts/etcd/values.yaml
```diff
@@ -19,4 +19,20 @@ resources:
...
     memory: 512Mi
-
+backup:
+  image:
+    repository: sapcc/etcdbrctl
+    tag: 0.3.0
+    pullPolicy: IfNotPresent
+  config:
+    schedule: "*/5 * * * *"
+    maxBackups: 10
```
databus23: Please add some comment inline what these values mean

jknipper: @databus23 I added comments and changed default values. Backups are kept for one week now.

pkg/util/etcd/etcd.go
jknipper: @databus23 I changed this.

pkg/client/openstack/admin/client.go
jknipper: @databus23 I changed this.

jknipper:
@databus23 I changed backup path and removed WIP status.

jknipper:
@databus23 I added a migration for storage container creation. Please have a look.

pkg/migration/3_etcdbr_create_storage_container.go
```diff
@@ -0,0 +1,29 @@
...
+	etcd_util "github.com/sapcc/kubernikus/pkg/util/etcd"
+)
+
+func CreateEtcdBackupStorageContainer(rawKluster []byte, current *v1.Kluster, client kubernetes.Interface, adminClient admin.AdminClient) (err error) {
+	secret, err := client.CoreV1().Secrets(current.GetNamespace()).Get(current.GetName(), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	username := string(secret.Data["openstack-username"])
```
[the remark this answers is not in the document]
jknipper: @databus23 I changed this accordingly.

pkg/migration/migration.go
```diff
@@ -17,7 +18,7 @@ var defaultRegistry Registry
 // The function is expected to modify the kluster accordingly, changed object is persisted
 // automatically after the handler returns with no error.
 // The kubernetes client can be used to modify other things (e.g. kluster secret)
-type Migration func(klusterRaw []byte, kluster *v1.Kluster, client kubernetes.Interface) (err error)
+type Migration func(klusterRaw []byte, kluster *v1.Kluster, client kubernetes.Interface, admin_client admin.AdminClient) (err error)
```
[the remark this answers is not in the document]
jknipper: @databus23 I changed this accordingly.

jknipper:
@databus23 I added e2e test, please have a look.

PR #329, opened 2018-10-30
Fix local traffic only services

databus23:
This PR ~hopefully~ fixes services with `externalTrafficPolicy=Local` in our setup.

See https://kubernetes.io/docs/setup/independent/troubleshooting-kubeadm/#services-with-externaltrafficpolicy-local-are-not-reachable for more details.

I'm waiting on positive feedback from the user that initially reported this problem that this change fixes the problem.

Note: I removed some trailing whitespaces in this PR as well. Look at the individual commits to see the actual changes.

### DIFF
--- a/pkg/templates/node_1.10.go
+++ b/pkg/templates/node_1.10.go
@@ -441,7 +441,7 @@ storage:
           enableProfiling: false
           featureGates: {}
           healthzBindAddress: 0.0.0.0:10256
-          hostnameOverride: ""
+          hostnameOverride: {{ .NodeName }}
           iptables:
             masqueradeAll: false
             masqueradeBit: 14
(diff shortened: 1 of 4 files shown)


databus23:
User is giving a 👍 So this is good to go from my side.

databus23:
@BugRuger. See 
https://github.com/sapcc/kubernikus/pull/329/commits/cd27ac95ef5514e5c505ff2af4893747ac1a0d4e

But maybe we should centralize the node name setting on `/etc/kubernetes/environment` instead of littering it various config files? Wdyt?

BugRoger:
Ah, I see. Curious that there's even have a docs page for that fix. 🤔 

ENV and slurp it in with systemd. Hm. I'd keep it in the config files. The trend is to move the options from flags to config files. I faintly remember that kube-proxy spits out warnings that the flags are already deprecated.

PR #331, opened 2018-11-03
Fix etcd backup e2e test and pyrolisis

jknipper:
This PR fixes etcd backup/restore e2e test. It deletes etcd data and tests its recovery. Also some cleanup of swift objects is done in pyrolisis phase.


jknipper:
@databus23 I fixed environment variables for control plane kluster name, that should be it. Please have a look.

PR #332, opened 2018-11-16
Fix pyrolisis test, wait for objects are actually deleted

jknipper:
This fixes pyrolisis in E2E tests for some cases. It waits for swift objects get actually deleted before deleting storage container.

### DIFF
--- a/test/e2e/pyrolisis_test.go
+++ b/test/e2e/pyrolisis_test.go
@@ -98,10 +98,13 @@ func (p *PyrolisisTests) CleanupBackupStorageContainers(t *testing.T) {
 				require.NoError(t, err, "There should be no error while deleting object %s/%s", container, object)
 			}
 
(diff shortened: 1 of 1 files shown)


test/e2e/pyrolisis_test.go
```diff
@@ -91,7 +98,11 @@ func (p *PyrolisisTests) CleanupBackupStorageContainers(t *testing.T) {
 				require.NoError(t, err, "There should be no error while deleting object %s/%s", container, object)
 			}
 
-			_, err = containers.Delete(storageClient, container).Extract()
+			wait.PollImmediate(CleanupBackupContainerDeleteInterval, CleanupBackupContainerDeleteTimeout,
+				func() (bool, error) {
+					_, err = containers.Delete(storageClient, container).Extract()
```
databus23: Instead of retrying for every error I Ouwl like it better if we only retry for 409 replies

PR #339, opened 2018-11-27
Cleanup e2e test instances in pyrolisis

jknipper:
This PR cleans up instances that are left from previous failing E2E tests.


deps/deps.go
```diff
@@ -7,6 +7,7 @@ import (
 	_ "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/quotasets"
 	_ "github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	_ "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/quotasets"
+	_ "github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
```
databus23: This is not required as we use this already somewhere else. we only need to list stuff here is that is exclusively used in tests

jknipper: I removed this from PR.

Issue #303, opened 2018-08-06
Allow Deletion of Node Pools


BugRoger:
Currently Node Pools can't be deleted for yet unknown reasons. Investigate the problem.

BugRoger:
Reopen @onuryilmaz @databus23

BugRoger:
@onuryilmaz as discussed 😉

onuryilmaz:
Closing the issue, thanks @edda for UI part.

PR #341, opened 2018-11-27
Allow Deletion of Node Pools

onuryilmaz:
This PR solves #303 for Kubernikus API:

- Only allow the changing size of `nodePool` instances during the update
- Do not allow deletion of `nodePool` if the size is larger than 0
- If the size is 0 and `nodePool` is deleted, clear from `spec.nodePools` and `status.nodePools`

When this PR is merged, we can allow deletion of `nodePools` from Elektra UI. /cc @edda

### DIFF
--- a/pkg/api/handlers/util.go
+++ b/pkg/api/handlers/util.go
@@ -1,10 +1,8 @@
 package handlers
 
 import (
-	"fmt"
-
 	"errors"
-
+	"fmt"
 	"net/http"
 	"strings"
 
--- a/pkg/api/handlers/util.go
+++ b/pkg/api/handlers/util.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"fmt"
+
 	"github.com/pkg/errors"
 
 	"net/http"
--- a/pkg/api/handlers/util.go
+++ b/pkg/api/handlers/util.go
@@ -3,7 +3,7 @@ package handlers
 import (
 	"fmt"
 
-	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


pkg/api/handlers/util.go
```diff
@@ -2,6 +2,9 @@ package handlers
 
 import (
 	"fmt"
+
+	"github.com/pkg/errors"
```
databus23: is the use of this package on purpose? I have a feeling this crept in `goimports` and the standard `errors` package should be used here.

onuryilmaz: Fixed, thanks 👍

pkg/api/handlers/util.go
```diff
@@ -2,6 +2,9 @@ package handlers
 
 import (
 	"fmt"
+
```
databus23: Excuse the nitpick. But can you remove the blank line here.

I try to enforce a pattern for imports to goes like this:

```
std library imports in one block
[blank line]
external imports in one block
[blank line]
github.com/sapcc/kubernikus imports in one block
```

Unfortunately goimports `-local` flag does not fully support this and allows for addition blank lines. See https://github.com/sapcc/kubernikus/blob/master/test/gofmt.sh#L9

pkg/api/handlers/util.go
```diff
@@ -2,6 +2,9 @@ package handlers
...
+	"errors"
+
```
databus23: this line as well

PR #359, opened 2019-01-03
Check for free space in object store on kluster creation

jknipper:
This PR checks for free space in object store on kluster creation, object store is needed for etcd backup. It also updates gophercloud to a recent version to get quota details.

### DIFF
--- a/pkg/util/etcd/etcd.go
+++ b/pkg/util/etcd/etcd.go
@@ -7,7 +7,8 @@ import (
 )
 
 const (
-	BackupStorageContainerBase = "kubernikus-etcd-backup"
+	BackupStorageContainerBase               = "kubernikus-etcd-backup"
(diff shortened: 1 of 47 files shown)


pkg/util/etcd/etcd.go
```diff
@@ -7,7 +7,8 @@ import (
 )
 
 const (
-	BackupStorageContainerBase = "kubernikus-etcd-backup"
+	BackupStorageContainerBase               = "kubernikus-etcd-backup"
+	BackupStorageContainerMinimumFreeStorage = 500000000 // 500MiB
```
databus23: I guess this will break all our soak test as we only have 100Mib of quota there if I'm not mistaken

jknipper: Yes, this is true. We can change this to 100MiB but this won't be enough in most cases.

PR #362, opened 2019-01-07
Export mount success events

jknipper:
This PR exports mount succes events to prometheus. It also renames metric labels to _total.

### DIFF
--- a/charts/kubernikus-system/vendor/eventexporter/values.yaml
+++ b/charts/kubernikus-system/vendor/eventexporter/values.yaml
@@ -8,7 +8,7 @@ metrics:
   config:
     config.yaml: |-
       metrics:
-      - name: volume_mount_error
+      - name: volume_mount_error_total
         event_matcher:
         - key: InvolvedObject.Kind
           expr: Pod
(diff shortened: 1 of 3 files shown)


charts/kubernikus-system/vendor/eventexporter/values.yaml
```diff
@@ -8,7 +8,7 @@ metrics:
   config:
     config.yaml: |-
       metrics:
-      - name: volume_mount_error
+      - name: volume_mount_error_total
```
BugRoger: Alerts will have to be updated when the metric names change.

charts/kubernikus-system/vendor/eventexporter/values.yaml
```diff
@@ -20,3 +20,13 @@ metrics:
           expr: Warning
         labels:
           node: Source.Host
+      - name: volume_mount_success_total
+        event_matcher:
+        - key: InvolvedObject.Kind
+          expr: Pod
+        - key: Message
+          expr: MountVolume.SetUp succeeded for volume .pvc-.*
```
BugRoger: missing `*` in the start?

PR #363, opened 2019-01-08
prometheus: update external_labels

databus23:
This change is depended on changes to the alertmanager

### DIFF
--- a/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
+++ b/charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
@@ -6,7 +6,10 @@ global:
   scrape_timeout: 55s
 
   external_labels:
-    region: k-{{ .Values.global.region }}
+    region: {{ required ".Values.global.region missing" .Values.global.region }}
+    cluster: {{ required ".Values.global.cluster missing" .Values.global.cluster }}
+    cluster_type: {{ required ".Values.global.cluster_type missing" .Values.global.cluster_type }}
+
 
 {{- if .Values.use_alertmanager }}
 alerting:
--- a/charts/kubernikus-system/charts/prometheus/kubernetes.alerts
+++ b/charts/kubernikus-system/charts/prometheus/kubernetes.alerts
@@ -105,7 +105,7 @@ groups:
       description: The pod {{ $labels.namespace }}/{{ $labels.pod_name }} killed several times in short succession. This could be due to wrong resource limits.
 
   - alert: VolumeAttachmentFailed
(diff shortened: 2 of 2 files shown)


charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -6,7 +6,10 @@ global:
   scrape_timeout: 55s
 
   external_labels:
-    region: k-{{ .Values.global.region }}
+    region: {{ required ".Values.global.region missing" .Values.global.region }}
+    cluster: {{ required ".Values.global.cluster missing" .Values.global.cluster }}
+    cluster_type: {{ required ".Values.global.cluster_type missing" .Values.global.cluster_type }}
```
BugRoger: What's a `cluster-type`? Also not used so far.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
```diff
@@ -6,7 +6,10 @@ global:
   scrape_timeout: 55s
 
   external_labels:
-    region: k-{{ .Values.global.region }}
+    region: {{ required ".Values.global.region missing" .Values.global.region }}
+    cluster: {{ required ".Values.global.cluster missing" .Values.global.cluster }}
```
BugRoger: This is not defined in the current values, or is it somewhere I haven't checked? Also, at least until now it doesn't seem to be used. 

Maybe better to remove it for the sake of this PR?

auhlig:
@BugRoger Regarding the `cluster`, `cluster_type` labels: https://github.com/sapcc/helm-charts/pull/524/commits/af347522ffe2e08565e5774454b51a4d6c68a454
Discussed with Fabus that `region` should always be equivalent to the openstack region - without `k-|s-|i-|..` prefix. The optional `cluster` label will be the name like `k-staging|k-master|k-something` etc. and since there can be more than one cluster of a kind in a region (scaleout, internet, kubernikus) the `cluster_type` helps to distinguish them.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
databus23: So the idea is that `cluster_type` contains something like `controlplane` or `kubernikus`. This is mainly useful to simply the alert routing allow alerts from different types to be routed differently. Without this we always need to enumerate all regions in the alert routes which is verbose and difficult to maintain.

charts/kubernikus-system/charts/prometheus/templates/_prometheus.yaml.tpl
databus23: There is a PR waiting in cc/secrets that contains the necessary values.

PR #364, opened 2019-01-08
Fixes etcd backup reauth and migrates existing klusters

jknipper:
This PR updates etcd backup/restore to current version 0.4.1 to fix token reauth bug. It also migrates existing klusters to new version.

### DIFF
--- a/charts/kube-master/charts/etcd/values.yaml
+++ b/charts/kube-master/charts/etcd/values.yaml
@@ -22,7 +22,7 @@ resources:
 backup:
   image:
     repository: sapcc/etcdbrctl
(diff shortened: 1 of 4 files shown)


charts/kube-master/charts/etcd/values.yaml
```diff
@@ -40,5 +40,5 @@ backup:
       cpu: 100m
       memory: 128Mi
     limits:
-      cpu: 300m
-      memory: 1Gi
+      cpu: 500m
+      memory: 1536Mi
```
BugRoger: Yikes, is this leaky or what's happening?

PR #368, opened 2019-01-15
Make kubelet container ready for rook ceph operator

jknipper:
This PR makes the kubelet in our core os nodes ready to run rook ceph operator/clusters.

### DIFF
--- a/pkg/templates/node_1.10.go
+++ b/pkg/templates/node_1.10.go
@@ -83,9 +83,13 @@ systemd:
           --volume var-lib-cni,kind=host,source=/var/lib/cni \
           --volume var-log,kind=host,source=/var/log \
           --volume etc-machine-id,kind=host,source=/etc/machine-id,readOnly=true \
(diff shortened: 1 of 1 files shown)


pkg/templates/node_1.10.go
```diff
@@ -83,9 +83,13 @@ systemd:
           --volume var-lib-cni,kind=host,source=/var/lib/cni \
           --volume var-log,kind=host,source=/var/log \
           --volume etc-machine-id,kind=host,source=/etc/machine-id,readOnly=true \
+          --volume modprobe,kind=host,source=/usr/sbin/modprobe \
+          --volume lib-modules,kind=host,source=/lib/modules \
```
databus23: I was under the impression this was already done In the Kubelet wrapper: https://github.com/coreos/coreos-overlay/blob/master/app-admin/kubelet-wrapper/files/kubelet-wrapper#L73

jknipper: @databus23 You are right, but modprobe is still needed for the operator. I changed this.

PR #369, opened 2019-01-18
Update eventexporter chart and config

jknipper:
This PR updates eventexporter which looks up pod data and discards old events. It also introduces resource limits.


charts/kubernikus-system/vendor/eventexporter/templates/deployment.yaml
```diff
@@ -24,14 +24,17 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         args:
+        - -discard=60
```
databus23: Can we make this a golang duration field? So that you can specify `60s` `5m` etc.

jknipper: Sure.

jknipper: @databus23 I changed this

PR #360, opened 2019-01-03
Run chart tests in travis make build

jknipper:
This PR runs helm chart template rendering and api version tests in travis.

PR #372, opened 2019-01-25
Add alert and chart testing to travis

jknipper:
This PR adds testing of prometheus alert configuration and chart rendering. It adds a docker image with all needed tools to run the tests. It is in favor of https://github.com/sapcc/kubernikus/pull/360


jknipper:
@databus23 I updated this, dockerfile is now in contrib. Please have a look.

PR #373, opened 2019-01-30
Use contrib tests image in e2e tests

jknipper:
Use contrib tests image in e2e tests instead of installing needed tools every time. This hopefully speeds up soak tests a bit.

### DIFF
--- a/ci/task_e2e_tests.yaml
+++ b/ci/task_e2e_tests.yaml
@@ -3,8 +3,8 @@ platform: 'linux'
 image_resource:
   type: docker-image
   source:
-    repository: golang
-    tag: 1.9.2-alpine3.6
+    repository: sapcc/kubernikus-tests
+    tag: latest
 
 inputs:
   - name: kubernikus.builds
@@ -17,11 +17,10 @@ run:
     - -c
     - |
       set -o pipefail
-      SENTRY_CLI=$PWD/cache/sentry-cli-1.37.0
+      SENTRY_CLI=/usr/local/bin/sentry-cli
       export GOPATH=$PWD/gopath
       cd gopath/src/github.com/sapcc/kubernikus
 
-      apk add --no-progress --no-cache make git curl
       export RUN_PARALLEL=false
       make test-e2e | tee test.output
       rc=$?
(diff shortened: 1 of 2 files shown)


ci/task_e2e_tests.yaml
```diff
@@ -17,20 +17,17 @@ run:
     - -c
     - |
       set -o pipefail
-      SENTRY_CLI=$PWD/cache/sentry-cli-1.37.0
+      SENTRY_CLI=/usr/local/bin/sentry-cli
```
databus23: If we do it like this then we can also remove the `cache` stanza from the task definition

ci/task_e2e_tests.yaml
```diff
@@ -3,8 +3,8 @@ platform: 'linux'
 image_resource:
   type: docker-image
   source:
-    repository: golang
-    tag: 1.9.2-alpine3.6
+    repository: sapcc/kubernikus-tests
+    tag: latest
```
databus23: Maybe this should be a specific version so that we are sure what is running...

jknipper:
@databus23 I removed cache path and added specific test image.

contrib/kubernikus-tests/Dockerfile
```diff
@@ -5,12 +5,13 @@ WORKDIR /go/src/github.com/sapcc/kubernikus/
 RUN apk add --no-cache make git curl bash
 
 RUN curl -Lf https://storage.googleapis.com/kubernetes-helm/helm-v2.10.0-linux-amd64.tar.gz \
-		| tar --strip-components=1 -C /usr/local/bin -zxv \
-		&& helm version -c
```
databus23: Why remove the binary invocation. The purpose of those are to check that download/upack indeed moves a valid binary in place

jknipper: @databus23 True, i added it to the other binaries as well.

ci/task_e2e_tests.yaml
jknipper: @databus23 I changed this.

ci/task_e2e_tests.yaml
jknipper: @databus23 I changed this.

PR #375, opened 2019-02-05
node pool name validation

### LINKED ISSUE #370: Tighten Node-Pool Name Validations
The node-pool name becomes part of the node's name. Currently we still allow underscores which leads to the following error:

```
kubelet_node_status.go:117] Unable to register node "lala-m2_xlarge-mh5gr" with API server: Node "lala-m2_xlarge-mh5gr" is invalid: metadata.name: Invalid value: "lala-m2_xlarge-mh5gr": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

Tighten the validations to conform with the regex Kubernetes uses.

onuryilmaz:
fixes #370 

* `go-swagger `version increased to v0.18.0 (latest version now)
  * There are issues opened for 0.13.0 about the validation of array items 
* server and clients are re-generated 
* vendor is updated with the requirements from v0.18.0

* When the node pool name is not valid, API server returns:
```
422 Unprocessable Entity
{
    "code": 605,
    "message": "name in body should match '^[a-z]([a-z0-9]*)?$'"
}
```


onuryilmaz:
TODO:

- [ ] limit node pool length
- [ ] check with Kubernetes node name restriction
- [ ] check with nova node name restriction

PR #378, opened 2019-02-14
Backward compatibility for node pool name validation

onuryilmaz:
This PR allows the following in addition to be backward compatible:
- digits as the first character in node pool name
- "." and "-" in the node pool name


databus23:
Thanks!

PR #380, opened 2019-02-14
adds availability zone aware node-pools

### LINKED ISSUE #258: Allow availability zone to be specified per nodepool
Long standing customer request is to be able to choose the availability zone per nodepool.
Effectively spreading the load.

This is barely usable as the kube-master is currently only in one AZ but it is a low hanging fruit.

BugRoger:
This PR adds functionality to explicitly specify an availability zone for node pools. 

It includes:
- Migration of existing node pools. Defaults are not queryable via API. Therefore this migration needs to be hardcoded. It is unfortunately CCloudEE specific. ☹️ We can turn this migration into a NOOP once it is rolled out.
- Makes `availabilityZone` a required field on node pools
- Disallows changes to the `availabilityZone` via API
- Creates additional StorageClasses `cinder-zone-a`, `cinder-zone-b`, ... Keeps `cinder-default`. This will choose an AVZ randomly of all existing nodes.

Needs to be released in lock-step with Elektra due to the now required `availabilityZone` field.

Fixes #258


pkg/migration/5_insert_avz_into_nodepools.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/sapcc/kubernikus/pkg/client/openstack"
+)
+
+func InsertAVZIntoNodePools(rawKluster []byte, current *v1.Kluster, client kubernetes.Interface, openstackFactory openstack.SharedOpenstackClientFactory) (err error) {
+	apiSecret, err := client.CoreV1().Secrets(current.Namespace).Get(fmt.Sprintf("%s-secret", current.Name), meta_v1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	secret, err := v1.NewSecret(apiSecret)
```
databus23: This can be simplified to use `secret, err := util.KlusterSecret(current)`

BugRoger: Optimized!

databus23:
Awesome work! Merge at will

PR #381, opened 2019-02-15
Restore API Functionality by Adding Nodepool AVZ Validations and Defaults 

BugRoger:
This PR restores the default behaviour of the `clusterCreate` API. It can now still be called without a required `availabilityZone` on the node pool.

A default AVZ is selected if non is provided. If an AVZ is provided it is validated against currently enabled AVZs.

### DIFF
--- a/pkg/api/handlers/create_cluster.go
+++ b/pkg/api/handlers/create_cluster.go
@@ -39,7 +39,7 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal 
 
 	metadata, err := fetchOpenstackMetadata(params.HTTPRequest, principal)
 	if err != nil {
-		return NewErrorResponse(&operations.CreateClusterDefault{}, 500, err.Error())
+		return NewErrorResponse(&operations.CreateClusterDefault{}, 500, "Failed to get OpenStack metadata: %s", err)
 	}
 
 	spec.Name = name
--- a/pkg/api/handlers/create_cluster.go
+++ b/pkg/api/handlers/create_cluster.go
@@ -58,7 +58,7 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal 
 			spec.NodePools[i].AvailabilityZone = avz
 		} else {
 			if err := validateAavailabilityZone(pool.AvailabilityZone, metadata); err != nil {
-				return NewErrorResponse(&operations.CreateClusterDefault{}, 409, "Availability Zone %s is invalid: %s", pool.AvailabilityZone, err)
+				return NewErrorResponse(&operations.CreateClusterDefault{}, 409, "Availability zone '%s' for pool '%s' is invalid: %s", pool.AvailabilityZone, pool.Name, err)
 			}
 		}
 	}
--- a/pkg/api/handlers/create_cluster.go
+++ b/pkg/api/handlers/create_cluster.go
@@ -37,11 +37,6 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal 
 		return NewErrorResponse(&operations.CreateClusterDefault{}, int(err.Code()), err.Error())
 	}
 
-	metadata, err := fetchOpenstackMetadata(params.HTTPRequest, principal)
-	if err != nil {
-		return NewErrorResponse(&operations.CreateClusterDefault{}, 500, "Failed to get OpenStack metadata: %s", err)
-	}
-
 	var metadata *models.OpenstackMetadata
 	spec.Name = name
 	for i, pool := range spec.NodePools {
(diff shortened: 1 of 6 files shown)


pkg/api/handlers/create_cluster.go
```diff
@@ -37,13 +37,32 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal
...
+		// Set default AvailabilityZone
+		if pool.AvailabilityZone == "" {
+			avz, err := getDefaultAvailabilityZone(metadata)
+			if err != nil {
+				return NewErrorResponse(&operations.CreateClusterDefault{}, 500, err.Error())
+			}
+			spec.NodePools[i].AvailabilityZone = avz
+		} else {
+			if err := validateAavailabilityZone(pool.AvailabilityZone, metadata); err != nil {
+				return NewErrorResponse(&operations.CreateClusterDefault{}, 409, "Availability Zone %s is invalid: %s", pool.AvailabilityZone, err)
```
databus23:
```suggestion
				return NewErrorResponse(&operations.CreateClusterDefault{}, 409, "Availability zone '%s' for pool '%s' is invalid: %s", pool.AvailabilityZone, pool.Name, err)
```

pkg/api/handlers/create_cluster.go
```diff
@@ -37,13 +37,32 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal
...
+		// Set default image
 		if pool.Image == "" {
 			spec.NodePools[i].Image = DEFAULT_IMAGE
 		}
+
+		// Set default AvailabilityZone
+		if pool.AvailabilityZone == "" {
+			avz, err := getDefaultAvailabilityZone(metadata)
+			if err != nil {
+				return NewErrorResponse(&operations.CreateClusterDefault{}, 500, err.Error())
```
databus23:
```suggestion
				return NewErrorResponse(&operations.CreateClusterDefault{}, 500, "Failed to get default availability zones: %s", err)
```

pkg/api/handlers/create_cluster.go
```diff
@@ -37,13 +37,32 @@ func (d *createCluster) Handle(params operations.CreateClusterParams, principal
 		return NewErrorResponse(&operations.CreateClusterDefault{}, int(err.Code()), err.Error())
 	}
 
+	metadata, err := fetchOpenstackMetadata(params.HTTPRequest, principal)
+	if err != nil {
+		return NewErrorResponse(&operations.CreateClusterDefault{}, 500, err.Error())
```
databus23:
```suggestion
		return NewErrorResponse(&operations.CreateClusterDefault{}, 500, "Failed to get OpenStack metadata: %s", err)
```

pkg/api/handlers/util.go
```diff
@@ -109,3 +112,42 @@ func nodePoolEqualsWithScaling(old, new models.NodePool) error {
...
+	authOptions := &tokens.AuthOptions{
+		IdentityEndpoint: principal.AuthURL,
+		TokenID:          tokenID,
+		AllowReauth:      true,
+		Scope: tokens.Scope{
+			ProjectID: principal.Account,
+		},
+	}
+
+	client, err := openstack.NewSharedOpenstackClientFactory(nil, nil, nil, getTracingLogger(request)).ProjectClientFor(authOptions)
```
databus23: This will break. We [cache the client](https://github.com/sapcc/kubernikus/blob/master/pkg/client/openstack/factory.go#L125) based on the project id (which is stupid in the first place because there might be more than one cluster eventually). But this means the apiserver will keep the client around reusing the first token and ignore the token that was passed in the current reques, once the token is expired this call will fail until the process is restarted.

pkg/api/handlers/util.go
```diff
@@ -109,3 +112,42 @@ func nodePoolEqualsWithScaling(old, new models.NodePool) error {
...
 	return nil
 }
+
+func fetchOpenstackMetadata(request *http.Request, principal *models.Principal) (*models.OpenstackMetadata, error) {
+	tokenID := request.Header.Get("X-Auth-Token")
+
+	authOptions := &tokens.AuthOptions{
+		IdentityEndpoint: principal.AuthURL,
+		TokenID:          tokenID,
+		AllowReauth:      true,
```
databus23: Reauth doesn't work for token based auth, once the token expires its game over.

BugRoger: removed the `AllowReauth`

pkg/api/handlers/util.go
BugRoger: The factory is recreated for each request. Ugly but the cache is only short-lived. 🤷‍♂️

databus23: I see. Ok so we can leave it like this for now. 👍

PR #382, opened 2019-02-18
Bump etcdbrctl to version 0.5.0

jknipper:
This PR updates etcd backup restore sidecar to version 0.5.0, this fixes a memory leak. It also changes resource limits to gardener defaults.


databus23:
Did you cross check if they did any changes to the Etcd eployment/statefulset/bootstrap.sh for this version?

jknipper:
@databus23 They removed the part where they moved etcd data to a subdirectory, rest is the same. I would like to keep it for update purposes. I tested it on a kluster and etcd and sidecar came up.

PR #383, opened 2019-02-18
e2e: Improve building black sameness checks

databus23:
* only check vms from the current kluster
* Print out instance ids in case of bb mismatch

### DIFF
--- a/test/e2e/node_test.go
+++ b/test/e2e/node_test.go
@@ -14,7 +14,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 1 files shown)


test/e2e/node_test.go
```diff
@@ -233,17 +233,20 @@ func (k *NodeTests) SameBuildingBlock(t *testing.T) {
...
 	bb := ""
-	for _, bbs := range s {
+	for i, bbs := range s {
 		require.NotEmpty(t, bbs.BuildingBlock, "Node building block should not be empty")
 		if bb == "" {
-			bb = string(bbs.BuildingBlock)
+			bb = bbs.BuildingBlock
 		} else {
-			require.Equal(t, bb, bbs.BuildingBlock, "Nodes should be on the same building block")
+			require.Equal(t, bb, bbs.BuildingBlock, "Node %s is on building block %s which is different from node %s which is running on %s", bbs.ID, bbs.BuildingBlock, s[i-1].ID, s[i-1].BuildingBlock)
```
jknipper: I we have more than two nodes it is not necessarily the previous one, isn't it? I think you should compare to the first one.

databus23: I could do that, sure. But the way the loop/check is designed it will break/fail as soon as the first different building block is found, so previous one will always be different. But your suggestion is a little bit simpler

PR #387, opened 2019-02-20
Seed Storageclasses

BugRoger:
This PR adds a migration that reseeds the storageClasses to already existing clusters.

Depends on #386


BugRoger:
trigger Travis with Close and Reopne

PR #399, opened 2019-03-20
Change kublet startup options if calico daemonset is deployed

databus23:
This is a quick hack that allows us to selectively run calico in `policy-only` mode to support network policies without changing the kubernikus api.

### DIFF
--- a/pkg/templates/ignition.go
+++ b/pkg/templates/ignition.go
@@ -40,7 +40,7 @@ func (i *ignition) getIgnitionTemplate(kluster *kubernikusv1.Kluster) string {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/templates/ignition.go
```diff
@@ -40,7 +40,7 @@ func (i *ignition) getIgnitionTemplate(kluster *kubernikusv1.Kluster) string {
 	}
 }
 
-func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, pool *models.NodePool, nodeName string, secret *kubernikusv1.Secret, imageRegistry version.ImageRegistry, logger log.Logger) ([]byte, error) {
+func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, pool *models.NodePool, nodeName string, secret *kubernikusv1.Secret, calicoNetworking bool, imageRegistry version.ImageRegistry, logger log.Logger) ([]byte, error) {
```
BugRoger: This method. 🤢

databus23: at least the new parameter will go again once we introduce proper "optional" features that are represented in the `Kluster` object.

PR #400, opened 2019-03-21
update vice president

auhlig:
also includes the renewed symantec sso cert for the vice api (via secrets).
tested manually in k-qa-de-1


auhlig:
rebased

PR #403, opened 2019-03-25
Add multus-cni and dedup-bridge to hyperkube cni plugins

SchwarzM:
This adds multus and dedup-bridge to the hyperkube image.

### DIFF
--- a/contrib/hyperkube/Makefile
+++ b/contrib/hyperkube/Makefile
@@ -3,6 +3,8 @@ IMAGE?=sapcc/hyperkube
 ARCH=amd64
 CNI_RELEASE=v0.6.0
 CNI_PLUGINS_RELEASE=v0.7.1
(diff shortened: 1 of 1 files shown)


contrib/hyperkube/Makefile
```diff
@@ -3,6 +3,8 @@ IMAGE?=sapcc/hyperkube
 ARCH=amd64
 CNI_RELEASE=v0.6.0
 CNI_PLUGINS_RELEASE=v0.7.1
+CNI_DEDUP_BRIDGE_RELEASE=v0.1.0
+MULTUS_RELEASE=v3.1
```
databus23: no 3.2?

SchwarzM: yes i want to replicate what we have installed prior to upgrading to unknown

PR #405, opened 2019-03-27
Adds Servicing Controller

BugRoger:
This controller handles node upgrades when the Kubernetes or CoreOS versions
are changed. It gracefully drains nodes before performing any action.

For Kubernetes upgrades the strategy is to replace the fleet by terminating
the nodes. CoreOS updates are handled by a soft reboot.

In order to allow the payload to settle only a single node per cluster is
processed at a time. Inbetween updates there's a 1h grace period.

In case any node in the cluster is unhealthy the upgrades are skipped. This
is to safeguard against failed upgrades destroying the universe.

For rollout and testing purposed the node upgrades are disabled by default.
They can manually be enabled by setting the node annotaion:

   kubernikus.cloud.sap/servicing=true


pkg/controller/servicing/lifecycler.go
```diff
@@ -0,0 +1,359 @@
...
+					"method":     "Reboot",
+				}).Add(1)
+		}
+	}(time.Now())
+	return lc.LifeCycler.Reboot(node)
+}
+
+// Replace collects metrics
+func (lc *InstrumentingLifeCycler) Replace(node *core_v1.Node) (err error) {
+	defer func(begin time.Time) {
```
databus23: the `prometheus.Labels` struct could be defined once in the beginning and passed to each `With` method

pkg/controller/servicing/lifecycler.go
```diff
@@ -0,0 +1,359 @@
...
+			events.FailedReplaceNode,
+			"Replacing node for upgrade: %v. Termination failed: %v",
+			node.GetName(),
+			err)
+	}
+	return err
+}
+
+// Drain collects metrics
+func (lc *InstrumentingLifeCycler) Drain(node *core_v1.Node) (err error) {
```
databus23: the `prometheus.Labels` struct could be defined once in the beginning and passed to each `With` method. maybe even in the `InstrumentingLifeCycler` struct and re-used in all the instrumenting functions.

pkg/controller/servicing/lister.go
```diff
@@ -0,0 +1,300 @@
...
+			"err", err,
+		)
+		return found
+	}
+
+	for _, node := range upgradable {
+		kubeletVersion, err := getKubeletVersion(node)
+		if err != nil {
+			d.Logger.Log(
+				"msg", "Couldn't get Kubelet version from Node. Skipping node upgrade.",
```
databus23: maybe add the node as a label here

pkg/controller/servicing/lister.go
```diff
@@ -0,0 +1,300 @@
...
+
+		if kubeletVersion.LessThan(klusterVersion) {
+			found = append(found, node)
+			continue
+		}
+
+		kubeProxyVersion, err := getKubeProxyVersion(node)
+		if err != nil {
+			d.Logger.Log(
+				"msg", "Couldn't get KubeProxy version from Node. Skipping node upgrade.",
```
databus23: same here, a node label could be useful

pkg/controller/servicing/reconciler.go
```diff
@@ -0,0 +1,203 @@
+package servicing
+
+import (
+	"time"
+
+	"k8s.io/client-go/tools/record"
+
```
databus23: a lot of unnecessary whitespace here...

swagger.yml
```diff
@@ -475,10 +475,14 @@ definitions:
     type: object
     x-nullable: false
     properties:
-      upgrade:
+      allowReboot:
```
databus23: a `description` field for the two fields would show up in the documentation and maybe safe as a few support inquires. 
In both cases its unclear what these fields actually do.

BugRoger: thanks. added

pkg/controller/servicing/lifecycler.go
```diff
@@ -173,15 +178,53 @@ func (lc *NodeLifeCycler) Replace(node *core_v1.Node) error {
...
+	copy := node.DeepCopy()
+	delete(copy.Annotations, AnnotationUpdateTimestamp)
+	err := lc.patch(node, copy)
+	if err != nil {
+		errors.Wrap(err, "failed to remove updating annotation")
+	}
+	return nil
+}
+
+func (lc *NodeLifeCycler) patch(old, new *core_v1.Node) error {
```
databus23: This is fairly complicated.
I thought doing it more straightforward like this:
```
patch := []byte(`[{"op":"remove","path":"/metadata/annotations/testing"}]`)
pod, err = client.Pods(namespace).Patch(pod.Name, types.JSONPatchType,
                patch)
```

BugRoger: Tell me about it... 🤣 

Now using `JSONPatchType`.

PR #406, opened 2019-03-29
randomize hourly etcd backup schedule

databus23:
This PR spreads our hourly full backups of etcd accross the entire hour by calculating a determintic value between 0-59 for each cluster based on the uid.


jknipper:
Looks good!

PR #408, opened 2019-04-01
Cluster upgrade

databus23:
This PR updates `groundctl` to perform Kubernetes version upgrades when `Spec.Version` is changed on the cluster object.

For now only manually triggered updates (by bumping Speed.Version) are supported. We'll ship automatic updates when the `kube-master` is bumped in a separate change and once very old clusters are upgraded once and subsequent changes are fairly small.

The way the upgrade mechanism works is by comparing `Spec.Version` and `Status.ApiserverVersion` and performing a helm upgrade when both differ. Once an upgrade is triggered the kluster change into the newly added`Upgrading` phase. In that phase no futher upgrade are allowed and all pods have to becoming running again and `Spec.Version` and `Status.ApiserverVersion` need to match to transition back into `Running` phase.

For now the upgrades are disabled by default and need to be explicitly enabled by adding the `kubernikus.cloud.sap/upgrade=yes` annotation to a cluster.

I also added a few migrations/changes to fixup older clusters so that they can be upgraded:
* Ensure the `Spec.OpenStack.SecurituGroupName` is not empty
* Ensure `Spec.Openstack.LBFloatingNetworkID` is not empty 
* Set `Spec.Version` to `Status.ApiserverVersion` to avoid downgrading manually upgraded clusters

While manually testing the upgrade I slipped in a few unrelated changes:

* Add a `kubernikus.cloud.sap/hammertime=no` annotation to disable the hammer time controller for a specific cluster
* Removed the `Status.Message` field from the `Kluster` object, we weren't really using it and not updating it properly


databus23:
This is ready for review.

PR #409, opened 2019-04-03
Disable etcdbr readiness check, expose metrics

jknipper:
This PR removes etcdbr readiness check and replaces it with testing etcd. This will ensure that etcd keeps running even if storage backend is failing. It also exposes metrics on backup status and alerts if backup is failing.

### DIFF
--- a/charts/kube-master/charts/etcd/templates/deployment.yaml
+++ b/charts/kube-master/charts/etcd/templates/deployment.yaml
@@ -66,6 +66,9 @@ spec:
       labels:
         app: {{ include "fullname" . }}
         release: {{ .Release.Name }}
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "8080"
     spec:
       volumes:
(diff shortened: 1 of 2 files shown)


jknipper:
@databus23 I added an alert, please have a look.

charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -100,16 +103,16 @@ spec:
...
           readinessProbe:
-            httpGet:
-              path: /healthz
-              port: 8080
+            exec:
+              command:
+                - /bin/sh
+                - -ec
+                - ETCDCTL_API=3 etcdctl get foo
             initialDelaySeconds: 15
```
databus23: I would lower this to 5 seconds, no need to artificially stall the readiness of a starting etcd pod

jknipper: Changed this.

PR #418, opened 2019-04-17
Change etcd backup policy to exponential

jknipper:
This PR changes etcdbr policy to exponential.


charts/kube-master/charts/etcd/values.yaml
```diff
@@ -27,14 +27,12 @@ backup:
     pullPolicy: IfNotPresent
   # do a full-backup every hour
   schedule: "15 * * * *"
-  # keep backups for one week
-  maxBackups: 168
```
databus23: if we remove this setting we can also remove `garbageCollectionPolicy` and hardcode it to exponential in the spec. Or we keep this setting which would allow to change the policy based on settings.

PR #420, opened 2019-04-23
[charts/kubernikus-system] remove vendor; use sapcc chart repository

auhlig:
Instead of copying  the charts we also use in the controlplane we could use the recently introduced sapcc helm chart repository to avoid diverging and outdated charts.

- Switch vendored charts to sapcc chart repository:
    - vice-president
    - oomkill-exporter
    - eventexporter

- While I was at it: bump kube-state-metrics, node-exporter to the versions we use in the controlplane.

EDIT: Tested in `k-qa-de-1`


auhlig:
**Note** that there's secrets PR#1083 that needs to be merged.

auhlig:
It's failing as travis can't reach our chart repository. 
We could make it available to the internet since the charts are already on public github

PR #422, opened 2019-04-25
Prometheus via operator

auhlig:
This PR dis-entangles the monitoring components and kubernikus-system chart. 
Introduces `kubernikus-monitoring` chart which contains:
- operated prometheus: prometheus operator + prometheus server base chart + custom config
- re-use of controlplane kubernetes alert- and aggregation rules from helm-charts
- `kubernikus-rules`: kubernikus alert- and aggregation rules as `PrometheusRule` CR
- move exporters(oomkill, event, etc.) from kubernikus-system to kubernikus-monitoring chart
- consume exporters from chart repository (same as in helm-charts) and remove from vendor

Extended pipeline accordingly.


auhlig:
Update: We avoid pulling kubelet, cadvisor metrics via proxy and scrape them directly from the node using its IP. Via http for kubelet metrics on `$nodeIP:4194/metrics`. Cadvisor metrics need https (not exposed via http) and `insecure_skip_verify: true` (cert only valid for nodeName) on `$nodeIP:10250/metrics`. Are you ok with this @databus23?

auhlig:
This is ready for a final review @databus23.

auhlig:
Rebased and incorporated comments @databus23.

This is good to merge from my side. 
Just want to manually trigger the deployment at least in k-qa-de-1 and supervise the other regions.

auhlig:
```
kind: Prometheus apiVersion: monitoring.coreos.com/v1 not matching minimum version requirements (1.7)!
```
😭 😭

databus23:
where are you seeing this. all k- clusters are running 1.11 now

auhlig:
It's "just" the chart test that doesn't like it: https://github.com/sapcc/kubernikus/blob/master/test/charts/charts.sh#L69

databus23:
I think we should be able to remove that check as all k-clusters have been upgraded to 1.11. This check was for safeguarding against resources that are not known to 1.7.

charts/kubernikus-monitoring/templates/_prometheus.yaml.tpl
```diff
@@ -215,48 +110,76 @@ scrape_configs:
...
+      action: replace
+      regex: ^/system\.slice/(.+)\.service$
+      target_label: systemd_service_name
+      replacement: '${1}'
+    - source_labels: [ id ]
+      action: replace
+      regex: ^/system\.slice/(.+)\.service$
+      target_label: container_name
+      replacement: '${1}'
+    - source_labels: [ namespace ]
```
databus23: Is this correct? That we drop everthing not running in `Kube-system`?

auhlig: Nope. Thanks for spotting this!

auhlig:
✅

test/charts/charts.sh
```diff
@@ -49,29 +50,5 @@ for chart in $pwd/charts/*; do
     fi
     cd ..
     echo "Done."
-
```
databus23: I think we should also be able to remove the allowed apiversions at the top of the file now

auhlig: done

PR #424, opened 2019-05-02
Bypass sni ingress proxy in the shared kubernetes client factory

databus23:
This should fix hanging node observeratory watches we had in the past one one if the control plane nodes goes away without warning.

We already use this for the controller-manger and scheduler were we didn't observe hanging client to my knowledge.


databus23:
@BugRoger ping

PR #426, opened 2019-05-03
Update k8sniff deployment

databus23:
This updates to k8sniff to our own build. Rebuilding the sniffer with no code changes with go 1.12.
This has a performance benefit as it now uses the `splice` syscall to proxy tcp connections.
And it also fixes hanging backend connections because go 1.12 enables tcp keepalives by default.

I'm also updateing the deployment to have 2 replicas running with anti affinity to improve availability of cluster apis in case of a control plane node failure


BugRoger:
Bam!

PR #427, opened 2019-05-07
Mitigate read-only root fs

SchwarzM:
Some kubernikus instances are currently hanging with read-only root fs due to failover in vmware.
The default timeout for the disk devices is 30 sec which is far too small for SAN which takes about 60 sec on average to failover.

This applies the udev rule vmware used to rollout with vmware-tools which has been superseeded by
open-vm-tools which does not ship with udev rules anymore.

### DIFF
--- a/pkg/templates/node_1.10.go
+++ b/pkg/templates/node_1.10.go
@@ -229,6 +229,19 @@ networkd:
 
 storage:
   files:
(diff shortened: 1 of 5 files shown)


databus23:
Where is this happening. I have never seen this?

BugRoger:
I tested it. It works for setting the timeout to 180.

PR #431, opened 2019-06-11
Add reject proxy sidecar and config

jknipper:
This PR adds MITM blocking proxy as a sidecar to the controller manager. The proxy is configured via a config map.

### DIFF
--- a/charts/kube-master/templates/cloud-controller-manager.yaml
+++ b/charts/kube-master/templates/cloud-controller-manager.yaml
@@ -1,5 +1,5 @@
-{{ if .Values.openstack }}
-{{ if (semverCompare ">= 1.13" .Values.version.kubernetes) }}
+{{- if .Values.openstack }}
+{{- if (semverCompare ">= 1.13" .Values.version.kubernetes) }}
 {{/* vim: set filetype=gotexttmpl: */ -}}
 apiVersion: "extensions/v1beta1"
 kind: Deployment
@@ -50,14 +50,14 @@ spec:
             items:
               - key: openstack.config
                 path: openstack.config
-        {{- if .Values.controllerManager.rejectProxy.enabled }}
+        {{- if .Values.rejectProxy.enabled }}
         - name: reject-cert-exchange
           emptyDir: {}
         {{- end }}
       initContainers:
-{{ if .Values.controllerManager.rejectProxy.enabled }}
+{{- if .Values.rejectProxy.enabled }}
         - name: generate-ca
-          image: "{{ .Values.controllerManager.rejectProxy.image.repository }}:{{ .Values.controllerManager.rejectProxy.image.tag }}"
+          image: "{{ .Values.rejectProxy.image.repository }}:{{ .Values.rejectProxy.image.tag }}"
           command:
             - /bin/sh
           args:
@@ -69,7 +69,8 @@ spec:
           volumeMounts:
             - mountPath: /tmp/reject-cert
               name: reject-cert-exchange
-{{ end }}
+      serviceAccountName: {{ include "master.fullname" . }}-reject-proxy
+{{- end }}
       containers:
         - name: cloud-controller-manager
           image: {{ include "cloudControllerManager.image" . | quote }}
@@ -78,10 +79,10 @@ spec:
           args:
             - -ce
             - |-
-{{ if .Values.controllerManager.rejectProxy.enabled }}
+{{- if .Values.rejectProxy.enabled }}
               cp /tmp/reject-cert/ca.crt /usr/local/share/ca-certificates/reject-proxy-ca.crt
               update-ca-certificates
-{{ end }}
+{{- end }}
               exec /bin/openstack-cloud-controller-manager \
               --allocate-node-cidrs=true \
               --authentication-kubeconfig=/etc/kubernetes/config/kubeconfig \
@@ -95,11 +96,11 @@ spec:
               --feature-gates=NodeLease=false \
 {{- end }}
               --kubeconfig=/etc/kubernetes/config/kubeconfig \
-{{ if eq (toString .Values.cloudControllerManager.replicaCount) "1" }}
+{{- if eq (toString .Values.cloudControllerManager.replicaCount) "1" }}
               --leader-elect=false \
 {{- end }}
               --use-service-account-credentials=true
-{{ if .Values.controllerManager.rejectProxy.enabled }}
+{{- if .Values.rejectProxy.enabled }}
           env:
             - name: http_proxy
               value: http://127.0.0.1:8080
@@ -107,7 +108,7 @@ spec:
               value: http://127.0.0.1:8080
             - name: no_proxy
               value: 127.0.0.1
-{{ end }}
+{{- end }}
           livenessProbe:
             httpGet:
               path: /healthz
@@ -125,13 +126,13 @@ spec:
             - mountPath: /etc/kubernetes/cloudprovider
               name: cloudprovider
               readOnly: true
-{{ if .Values.controllerManager.rejectProxy.enabled }}
+{{- if .Values.rejectProxy.enabled }}
             - mountPath: /tmp/reject-cert
               name: reject-cert-exchange
-{{ end }}
+{{- end }}
           resources:
 {{ toYaml .Values.cloudControllerManager.resources | indent 12 }}
-{{ if .Values.controllerManager.rejectProxy.enabled }}
+{{- if .Values.rejectProxy.enabled }}
         - name: reject-proxy
           command:
             - /bin/sh
@@ -143,7 +144,8 @@ spec:
               -namespace={{ .Release.Namespace }} \
               -cert-dir=/tmp/reject-cert \
               -config-map={{ include "master.fullname" . }}-reject-proxy
-          image: "{{ .Values.controllerManager.rejectProxy.image.repository }}:{{ .Values.controllerManager.rejectProxy.image.tag }}"
+          image: "{{ .Values.rejectProxy.image.repository }}:{{ .Values.rejectProxy.image.tag }}"
+          imagePullPolicy: {{ .Values.rejectProxy.image.pullPolicy }}
           ports:
             - containerPort: 8080
               name: reject-proxy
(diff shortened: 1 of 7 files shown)


charts/kube-master/templates/controller-manager.yaml
```diff
@@ -77,24 +79,38 @@ spec:
...
-            - --controllers=*,bootstrapsigner,tokencleaner
-            - --kubeconfig=/etc/kubernetes/config/kubeconfig
-            - --leader-elect=false
-            - --root-ca-file=/etc/kubernetes/certs/tls-ca.pem
-            - --service-account-private-key-file=/etc/kubernetes/certs/apiserver-clients-ca-key.pem
-            - --service-cluster-ip-range={{ .Values.serviceCIDR }}
-            - --use-service-account-credentials
+            - -ce
+            - |-
+              while ! wc /tmp/reject-cert/ca.crt; do sleep 1; done
```
databus23: I assume this is required because the cert gets generated by the sidecar. I would suggest to have an init container using the same image than the sidecar. This sidecar should generate the tickets during pod initialization (e.g. ` /reject-proxy -cert-dir=/tmp/reject-cert generate-ca`)
Using that its always clear that the cert has been generated when both containers (cm and sidecar) are starting.

charts/kube-master/templates/controller-manager.yaml
```diff
@@ -77,24 +79,38 @@ spec:
...
+              cp /tmp/reject-cert/ca.crt /usr/local/share/ca-certificates/reject-proxy-ca.crt
+              update-ca-certificates
+              /hyperkube \
```
databus23:
```suggestion
              exec/hyperkube \
```

charts/kube-master/templates/controller-manager.yaml
```diff
@@ -114,8 +130,31 @@ spec:
...
+              name: reject-cert-exchange
           resources:
 {{ toYaml .Values.controllerManager.resources | indent 12 }}
+        - name: reject-proxy
+          command:
+            - /reject-proxy
+            - -v=true
+            - -namespace=kubernikus
+            - -cert-dir=/tmp/reject-cert
+            image: hub.global.cloud.sap/d074427/reject-proxy:0.1.0
```
databus23: should be a chart value

jknipper:
@databus23 This feature can be now turned on/off and ca certificate is generated in init container. Please have a look.

databus23:
Where is the code for the reject proxy located? I would like to have a publicly available image on sapcc/reject-proxy or something.

databus23:
@jknipper if time permits could you rebase this PR and adopt changes for new kubernetes versions? the dedicated cloud-controleller-manager deployment is now responsibel for the loadbalancer managemend

jknipper:
@databus23 Code is now in https://github.com/sapcc/reject-proxy and image publicly available on docker hub `sapcc/reject-proxy`.

jknipper:
@databus23 I rebased on current master. Please have a look!

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -1,5 +1,5 @@
-{{- if .Values.openstack }}
-{{- if (semverCompare ">= 1.13" .Values.version.kubernetes) }}
```
databus23: Why remove those, while the first one does nothing, the second one removes a blank line at the top. I consider it good practise to have this in front of all "conditional" blocks to avoid spurious blank lines.

There are more instances of this below. I added suggestions for a few but stopped after a while to avoid annoying repetitions. I would suggest to add `{{-` infront of every `if` and corresponding `end` block.

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -50,27 +50,64 @@ spec:
...
+              --cluster-cidr={{ .Values.clusterCIDR }} \
+              --cluster-name={{ .Values.name }} \
+              --configure-cloud-routes=true \
 {{- if (semverCompare ">= 1.14" .Values.version.kubernetes) }}
-            - --feature-gates=NodeLease=false
+              --feature-gates=NodeLease=false \
 {{- end }}
-            - --kubeconfig=/etc/kubernetes/config/kubeconfig
+              --kubeconfig=/etc/kubernetes/config/kubeconfig \
 {{ if eq (toString .Values.cloudControllerManager.replicaCount) "1" }}
```
databus23:
```suggestion
{{- if eq (toString .Values.cloudControllerManager.replicaCount) "1" }}
```

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -50,27 +50,64 @@ spec:
...
-            - --leader-elect=false
+              --leader-elect=false \
 {{- end }}
-            - --use-service-account-credentials=true
+              --use-service-account-credentials=true
+{{ if .Values.controllerManager.rejectProxy.enabled }}
```
databus23:
```suggestion
{{- if .Values.controllerManager.rejectProxy.enabled }}
```

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -50,27 +50,64 @@ spec:
...
+          env:
+            - name: http_proxy
+              value: http://127.0.0.1:8080
+            - name: https_proxy
+              value: http://127.0.0.1:8080
+            - name: no_proxy
+              value: 127.0.0.1
+{{ end }}
```
databus23:
```suggestion
{{- end }}
```

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -88,7 +125,34 @@ spec:
             - mountPath: /etc/kubernetes/cloudprovider
               name: cloudprovider
               readOnly: true
+{{ if .Values.controllerManager.rejectProxy.enabled }}
```
databus23:
```suggestion
{{- if .Values.controllerManager.rejectProxy.enabled }}
```

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -88,7 +125,34 @@ spec:
...
+            - mountPath: /tmp/reject-cert
+              name: reject-cert-exchange
+{{ end }}
```
databus23:
```suggestion
{{- end }}
```

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -88,7 +125,34 @@ spec:
...
           resources:
 {{ toYaml .Values.cloudControllerManager.resources | indent 12 }}
-{{- end }}
-{{- end }}
+{{ if .Values.controllerManager.rejectProxy.enabled }}
```
databus23:
```suggestion
{{- if .Values.controllerManager.rejectProxy.enabled }}
```

charts/kube-master/templates/cloud-controller-manager.yaml
```diff
@@ -50,27 +50,64 @@ spec:
             items:
               - key: openstack.config
                 path: openstack.config
+        {{- if .Values.controllerManager.rejectProxy.enabled }}
```
databus23: Maybe we should remove the `controllerManager` part of this value path: `{{- if .Values.rejectProxy.enabled }}`

charts/kube-master/templates/reject-proxy-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "master.fullname" . }}-reject-proxy
+  namespace: {{ .Release.Namespace }}
+data:
+# Configuration example
+#  config: |-
+#    blockedUrls:
```
databus23: How about changing the config structure so that every block list entry is a separate ConfigMap entry. This makes dynamically updating the block list a lot simpler because individual entries can be targeted for addition/deletion.
e.g.:
```
data:
  neutron-lbaas: |
    method: POST
    url: https://someurl:443/v2.0/lbaas.*
    code:  int #should be optional defaulting to 403
    message: string #should be optional default to something generic
```

charts/kube-master/templates/cloud-controller-manager.yaml
jknipper: I removed this and changes values.

charts/kube-master/templates/reject-proxy-config.yaml
jknipper: I changed config format.

charts/kube-master/templates/controller-manager.yaml
jknipper: I changed this everywhere.

charts/kube-master/templates/cloud-controller-manager.yaml
jknipper: I changed this everywhere.

charts/kube-master/templates/reject-proxy-config.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if .Values.rejectProxy.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "master.fullname" . }}-reject-proxy
+  namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
```
databus23: I think this can be a `Role` and `RoleBinding`, no? The reject proxy only needs to access a ConfigMap in the same namespace as itself is running afais.

charts/kube-master/templates/reject-proxy-config.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if .Values.rejectProxy.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "master.fullname" . }}-reject-proxy
+  namespace: {{ .Release.Namespace }}
```
databus23: I would remove those references, the can added automatically and just add visual noise to the template

charts/kube-master/templates/reject-proxy-config.yaml
jknipper: I changed this.

charts/kube-master/templates/reject-proxy-config.yaml
jknipper: I removed namespace references.

PR #435, opened 2019-06-14
Add backup object expiration date

jknipper:
This PR adds expiration date to etcd backup storage objects on kluster deletion. For now objects get deleted after 30 days.

### DIFF
--- a/pkg/controller/ground.go
+++ b/pkg/controller/ground.go
@@ -592,6 +592,18 @@ func (op *GroundControl) terminateKluster(kluster *v1.Kluster) error {
 		domain := secret.Openstack.DomainName
 		//If the cluster was still in state Pending we don't have a service user yet: skip deletion
 		if username != "" && domain != "" {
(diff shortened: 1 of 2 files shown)


pkg/controller/ground.go
```diff
@@ -592,6 +592,18 @@ func (op *GroundControl) terminateKluster(kluster *v1.Kluster) error {
...
+				return fmt.Errorf("Couldn't get provider client for project: %s", err)
+			}
+			if err := etcd_util.SetObjectStorageExpiration(
+				projectProviderClient,
+				etcd_util.DefaultStorageContainer(kluster),
+				time.Duration(time.Hour*720), // delete after 30 days
+			); err != nil {
+				return fmt.Errorf("Couldn't set backup object store expiration: %s", err)
+			}
+
```
jknipper: Problem seems that actual deletion of kluster is happening after this, there are some backup leftovers. @databus23 Can we move deletion of service user after deletion of helm chart?

PR #438, opened 2019-06-20
Check prometheus.yaml

auhlig:
Rendering and validating `prometheus.yaml` beforehand. LGTY @databus23?
Tested locally


test/alerts/alerts.sh
```diff
@@ -8,3 +8,44 @@ if [ $? -ne 0 ]; then
     echo "Checking of prometheus rules failed."
     exit 1
 fi
+
+helm init --client-only
+helm repo add bugroger-charts https://raw.githubusercontent.com/BugRoger/charts/repo
```
databus23: do we still need this?  I thought this would be gone with the switch to stable grafana?

auhlig: needed until #439 is merged. removing repo in https://github.com/sapcc/kubernikus/pull/439/commits/f08882bdd00373288b27ce5e9805714b23c0f35f

auhlig:
removing repo also from this test. thus it must be merged after #439

PR #441, opened 2019-06-24
[WIP] Add metrics proxy and services

jknipper:
This PR adds metrics proxy to kubernikus-system and exposes metrics services of apiserver, etcd, controller manager and scheduler. It uses a Kubernetes API for authentication.


databus23:
TIL this feature is partly obsoleted by upstream kubernetes. With kubernetes 1.12 the controller-manager and scheduler get `--kubeconfig-authentication/authorization` flags to check RBAC permissions of requests to the metrics/debug port against RBAC rules. This means only the etcd metrics are left to be exposed by this proxy.

PR #447, opened 2019-08-01
add support for k8s 1.12

BugRoger:


BugRoger:
Manual regression tests. lgtm. Good for pre-release.

pkg/controller/ground/bootstrap/dns/manifests.go
```diff
@@ -143,8 +146,8 @@ spec:
         args:
         - --v=2
         - --logtostderr
-        - --probe=kubedns,127.0.0.1:10053,kubernetes.default.svc.{{ .Domain }},5,A
-        - --probe=dnsmasq,127.0.0.1:53,kubernetes.default.svc.{{ .Domain }},5,A
+        - --probe=kubedns,127.0.0.1:10053,kubernetes.default.svc.{{ .Domain }},5,SRV
```
databus23: Is this backward compatible and won't break cluster that are created with a version <1.12

BugRoger: That change is in since 2years/1.9.0. It's wierd that it works with the `A` record. According to the docs kube-dns creates a `SRV` entry for headless services, which the `kubernetes.default` should be. https://github.com/kubernetes/kubernetes/commit/3e4b9fad6a4babf0311ffca6aa438d04464bb71c

charts/kube-master/templates/controller-manager.yaml
```diff
@@ -90,6 +90,8 @@ spec:
             - --configure-cloud-routes=true
             - --controllers=*,bootstrapsigner,tokencleaner
             - --kubeconfig=/etc/kubernetes/config/kubeconfig
+            - --authentication-kubeconfig=/etc/kubernetes/config/kubeconfig
```
databus23: I think this flags should be conditional, no? Otherwise the deployment will break for older clusters

BugRoger: yes, thanks. I think, I'm closing this PR and rolling it into the v1.xx PR.

PR #449, opened 2019-08-01
adds support for k8s 1.12, 1.13, 1.14 and 1.15

BugRoger:
This PR adds support for Kubernetes 1.12, 1.13, 1.14 and 1.15.

Starting with v1.13 it is now using CoreDNS instead of kube-dns.
Existing clusters will stick with kube-dns.

Starting with v1.13 it is now also utlitizing the external OpenStack
cloud controller manager. Existing clusters will be upgraded to use the new
CCM. With the new CCM it is now required to set a cluster-name to avoid
name clashes with load-balancers.

Kubernikus used to default to "Kubernetes" as cluster name. With this PR
it is now changing the name to the kluster-name. Existing load-balancers
and volumes will continue to work (as of CCM v1.15.0) using the legancy
name. Newly created LBs will use the new naming scheme.

Starting with v1.14 Kubernetes introduced a new NodeLease feature to
lessen the load on big clusters. We disable the `NodeLease` feature as
it clashes with the Hammertime controller. To be enabled in a secondary
step that obsoletes Hammertime as well. Depends on update to client-go
1.14


BugRoger:
@databus23 Ready for review

pkg/controller/ground/bootstrap/dns/coredns.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	manifest := `
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: coredns
+  namespace: kube-system
+  labels:
+    k8s-app: kube-dns
```
databus23: Is the use of `kube-dns` in this deployment deliberate? Like meaning "the Kubernetes dns service" of is this a copy and paste error from the old `kube-dns` stack?

BugRoger: Nicely spotted. I saw it as well and wondered. It is aligned with the default deployment here: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/coredns/coredns.yaml.base#L87

My assumption is, that it relates to the `kube-dns` service. Maybe to smooth out an upgrade without interruption.

databus23:
Jiha!

PR #450, opened 2019-08-06
Use upstream Terraform with upstream providers

kayrus:
The following resources have to be renamed (terraform state modification)

* `openstack_networking_rbacpolicies_v2` -> `openstack_networking_rbac_policy_v2`
* `ccloud_quota` -> `ccloud_project_quota_v1`
* `ccloud_kubernetes` -> `ccloud_kubernetes_v1`
* The `region` parameter was renamed to `endpoint_region` parameter in `openstack_identity_endpoint_v3`

### DIFF
--- a/terraform/kubernikus.tf
+++ b/terraform/kubernikus.tf
@@ -9,7 +9,7 @@ provider "openstack" {
 }
 
 provider "openstack" {
-  alias            = "master"
+  alias = "master"
 
   auth_url         = "https://identity-3.${var.region}.cloud.sap/v3"
   region           = "${var.region}"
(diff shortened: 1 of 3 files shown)


terraform/kubernikus.tf
```diff
@@ -21,7 +21,7 @@ provider "openstack" {
 }
 
 provider "openstack" {
-  alias            = "master.na-us-1"
+  alias = "master.na-us-1"
```
BugRoger: From my experience Terraform 12 doesn't like `.`s in the alias scheme. There is a validation. Unfortunately, the validation also breaks the migration that upgrades the state from 11 to 12.

BugRoger:
We migrated the pipelines to Terraform 0.12. Thanks for the great work on the provider :)

The deployment pipelines moved into a seperate repository.

PR #452, opened 2019-08-07
Add secure etcd communication

jknipper:
This PR adds etcd certificates and configuration for secure communication with apiserver and backup sidecar. It is disabled by default for now.

### DIFF
--- a/charts/kube-master/charts/etcd/templates/deployment.yaml
+++ b/charts/kube-master/charts/etcd/templates/deployment.yaml
@@ -43,11 +43,27 @@ data:
     exec etcd \
       --name=kubernikus \
       --data-dir=/var/lib/etcd/new.etcd \
+{{- if .Values.secure.enabled }}
+      --cert-file=/etc/kubernetes/certs/tls-etcd.pem \
+      --key-file=/etc/kubernetes/certs/tls-etcd-key.pem \
+      --client-cert-auth=true \
+      --trusted-ca-file=/etc/kubernetes/certs/etcd-clients-ca.pem \
+      --peer-cert-file=/etc/kubernetes/certs/tls-etcd.pem \
+      --peer-key-file=/etc/kubernetes/certs/tls-etcd-key.pem \
+      --peer-client-cert-auth=true \
+      --peer-trusted-ca-file=/etc/kubernetes/certs/etcd-peers-ca.pem \
+      --advertise-client-urls=https://${ETCD_IP}:2379 \
+      --initial-advertise-peer-urls=https://${ETCD_IP}:2380 \
+      --initial-cluster=kubernikus=https://${ETCD_IP}:2380 \
+      --listen-client-urls=https://0.0.0.0:2379 \
+      --listen-peer-urls=https://${ETCD_IP}:2380
+{{- else }}
       --advertise-client-urls=http://${ETCD_IP}:2379 \
       --initial-advertise-peer-urls=http://${ETCD_IP}:2380 \
       --initial-cluster=kubernikus=http://${ETCD_IP}:2380 \
       --listen-client-urls=http://0.0.0.0:2379 \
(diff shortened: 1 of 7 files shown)


charts/kube-master/templates/api.yaml
```diff
@@ -165,6 +182,16 @@ spec:
...
+          livenessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - cat < /dev/null > /dev/tcp/{{ include "etcd.fullname" . }}/2379
+            initialDelaySeconds: 60
+            periodSeconds: 5
+            timeoutSeconds: 1
+            failureThreshold: 1
```
jknipper: @databus23 I also added a liveness probe to solve our cache problem if etcd is gone.

jknipper: Moved this to a separate PR.

charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -43,11 +43,27 @@ data:
     exec etcd \
       --name=kubernikus \
       --data-dir=/var/lib/etcd/new.etcd \
+{{- if .Values.secure.enabled }}
+      --cert-file=/etc/kubernetes/certs/tls-etcd.pem \
+      --key-file=/etc/kubernetes/certs/tls-etcd-key.pem \
+      --client-cert-auth=true \
+      --trusted-ca-file=/etc/kubernetes/certs/etcd-clients-ca.pem \
+      --peer-cert-file=/etc/kubernetes/certs/tls-etcd.pem \
```
BugRoger: Shouldn't this be a certificate signed with the `etcd-peers-ca`?

pkg/apis/kubernikus/v1/secret.go
```diff
@@ -103,6 +105,11 @@ type Certificates struct {
 	TLSWormholePrivateKey   string `json:"tls-wormhole-key.pem"`
 	TLSWormholeCertificate  string `json:"tls-wormhole.pem"`
 
+	TLSEtcdCAPrivateKey  string `json:"tls-etcd-ca-key.pem"`
```
BugRoger: What is Etcd using TLS for?

pkg/apis/kubernikus/v1/secret.go
```diff
@@ -103,6 +105,11 @@ type Certificates struct {
...
+	TLSEtcdCACertificate string `json:"tls-etcd-ca.pem"`
+	TLSEtcdPrivateKey    string `json:"tls-etcd-key.pem"`
```
BugRoger: Shouldn't this be a `EtcdPeersPrivateKey`? Do we even have etcd peers? 🤣

databus23:
go away stale bot

jknipper:
@databus23 I rebased this on current master, looks good afaik.

jknipper:
CA and cert generation has been moved to separate PRs #502 #515

charts/kube-master/charts/etcd/templates/deployment.yaml
databus23: @jknipper @BugRoger has a point here I think. we should use a different CA for peer communication. As we don't use more then one etcd I also think the whole peer configuration can be omitted, no?

jknipper:
@databus23 I removed all peer settings from deployment. Please have a look.

pkg/apis/kubernikus/v1/secret.go
databus23: This is the server ca for client taking to etcd. its fine. :)

pkg/apis/kubernikus/v1/secret.go
databus23: for encryption I guess.

databus23:
@BugRoger good to go from my side. from your end as well?

BugRoger:
LGTM

PR #453, opened 2019-08-08
Restart apiserver if etcd is gone

jknipper:
This PR adds a liveness probe to the apiserver to solve caching problems. It is now restarted if etcd is gone for some reason.

### DIFF
--- a/charts/kube-master/templates/api.yaml
+++ b/charts/kube-master/templates/api.yaml
@@ -100,6 +100,8 @@ spec:
             items:
               - key: local-kubeconfig
                 path: kubeconfig
+              - key: liveness-probe.sh
+                path: liveness-probe.sh
       initContainers:
         - name: etcd-wait
           image: "{{ required "etcd.image.repository undefined" .Values.etcd.image.repository }}:{{ required "etcd.image.tag undefined" .Values.etcd.image.tag }}"
@@ -181,15 +183,14 @@ spec:
             - mountPath: /etc/kubernetes/bootstrap
               name: bootstrap
               readOnly: true
+            - mountPath: /
+              name: liveness-probe.sh
+              readOnly: true
           livenessProbe:
-            httpGet:
-              path: /healthz
-              port: 6443
-              scheme: HTTPS
+            command:
+              - /liveness-probe.sh
             initialDelaySeconds: 60
-            periodSeconds: 5
-            timeoutSeconds: 1
-            failureThreshold: 1
+            periodSeconds: 60
           readinessProbe:
             httpGet:
               path: /healthz
--- a/charts/kube-master/templates/api.yaml
+++ b/charts/kube-master/templates/api.yaml
@@ -99,8 +99,8 @@ spec:
             name: {{ include "master.fullname" . }}
             defaultMode: 0356
             items:
-              - key: api-liveness-probe.sh
-                path: api-liveness-probe.sh
+              - key: api-liveness-probe.py
+                path: api-liveness-probe.py
         - name: wormhole-config
           configMap:
             name: {{ include "master.fullname" . }}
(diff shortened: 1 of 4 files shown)


databus23:
Can't we use this recommendation: https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/monitoring.md#health-check

> Since v3.3.0, in addition to responding to the /metrics endpoint, any locations specified by --listen-metrics-urls will also respond to the /health endpoint. This can be useful if the standard endpoint is configured with mutual (client) TLS authentication, but a load balancer or monitoring service still needs access to the health check.

This should allow us to use a regular https liveness probe against the `/healthz` endpoint

jknipper:
Looks like we can use `/healthz` of apiserver for liveness probe. It is checking availability of etcd as well.

charts/kube-master/templates/api.yaml
```diff
@@ -165,6 +165,15 @@ spec:
             - mountPath: /etc/kubernetes/bootstrap
               name: bootstrap
               readOnly: true
+          livenessProbe:
+            httpGet:
+              path: /healthz
```
databus23:
```suggestion
              path: /healthz/etcd
```

databus23: I think `healthz` aggregates various health checks, we can restrict killing the pod to when the etcd endpoint check is failing specifically.

charts/kube-master/templates/api.yaml
```diff
@@ -165,6 +165,15 @@ spec:
...
+              port: 6443
+              scheme: HTTPS
```
databus23: Does this work for the self generated cert here? Does the health check run with `insecure` by default?

databus23: seems like it does

databus23:
I second Michael concerns regarding how aggressive the test is in this PR. I would like it to be more  permissive. My hop would be that the outage in case of an etcd restoration is not only a few seconds but a minute or something.

charts/kube-master/templates/api.yaml
jknipper: Indeed, this is interesting. It also works for readiness probe, see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes

> If scheme field is set to HTTPS, the kubelet sends an HTTPS request skipping the certificate verification.

jknipper:
I made it that aggressive because restart after deletion of etcd pod takes roughly 12 seconds. I can make it more relaxed but i don't know if we see all etcd restarts then.

databus23:
12 seconds is tough. to get with this health check. I'm a little worried that the apiserver will get dos'ed in an overload situation.

jknipper:
@databus23 I changed the probe and now it crawls backup sidecar metrics and checks if etcd data has been restored from backup. Unfortunately it adds much more complexity. Please have a look!

databus23:
@jknipper thanks!

Thinking about it downloading something in a liveness probe is really dangerous. if the network is slow the liveness probe might timeout and cause a crash loop until the network situation is resolved.

I took another look at the container and there is `socat` installed which can be used to make a simple http request:

```
echo -e "GET /metrics HTTP/1.0\r\n" |socat tcp4-connect:100.65.0.174:8080 STDIO | grep 'etcdbr_restoration_duration_seconds_count{succeeded="true"}'
```

looking at the bash code that comes afterwards I also thought maybe it would also be an option to write the whole probe in python. as apt is installed so as a full python interpreter. writing it in a proper script language get be more reliable then dealing with also those bash failure cases.

jknipper:
@databus23 I replaced the probe with your python implementation.

BugRoger:
Nice hack. 

image

databus23:
Can you rebase once so we can have a clean merge commit?

PR #456, opened 2019-08-09
Add cert renewal controller

jknipper:
This PR adds a controller to renew expired certificates.

### DIFF
--- a/pkg/util/certificates.go
+++ b/pkg/util/certificates.go
@@ -106,13 +106,14 @@ func (ca Bundle) Sign(config Config) (*Bundle, error) {
 }
 
 type CertificateFactory struct {
-	kluster *v1.Kluster
-	store   *v1.Certificates
-	domain  string
+	kluster     *v1.Kluster
+	store       *v1.Certificates
+	domain      string
+	certUpdates []string
 }
 
 func NewCertificateFactory(kluster *v1.Kluster, store *v1.Certificates, domain string) *CertificateFactory {
-	return &CertificateFactory{kluster, store, domain}
+	return &CertificateFactory{kluster, store, domain, nil}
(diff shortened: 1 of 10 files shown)


pkg/util/certificates.go
```diff
@@ -123,6 +133,8 @@ func (cf *CertificateFactory) Ensure() error {
 		return fmt.Errorf("Failed to parse clusters advertise address: %s", cf.kluster.Spec.AdvertiseAddress)
 	}
 
+	cf.certUpdates = []CertUpdates{}
```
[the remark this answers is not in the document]
jknipper: @databus23 I changed this, updates are now returned right away.

pkg/util/certificates.go
```diff
@@ -108,91 +111,147 @@ type CertificateFactory struct {
...
+	Type        string
+	Reason      string
+}
+
 func NewCertificateFactory(kluster *v1.Kluster, store *v1.Certificates, domain string) *CertificateFactory {
 	return &CertificateFactory{kluster, store, domain}
 }
 
-func (cf *CertificateFactory) Ensure() error {
+func (cf *CertificateFactory) Ensure() (error, []CertUpdates) {
```
databus23: Thanks for taking care! I have one final remark, can we reverse the order of the return values? In general in go its custom that the error is always returned as the last argument. Its very uncommon to deviate from this.

jknipper: @databus23 I changed return order of `Ensure()`.

databus23: ❤️

databus23: Will merge as soon as the test is green. Thanks a lot

PR #459, opened 2019-08-16
adds cname operator aka disco

auhlig:
Disabled by default. Enabled via regional secrets. 
Ingresses need
```
metadata:
  annotations: 
    disco: "true"
```

Opening a PR to secrets.git proving the credentials.


auhlig:
not needed right now

PR #461, opened 2019-08-20
credential api: embed cluster version in cluster name

databus23:
I tried using the extensions fields that are defined in the Config api type but couldn't get them to work. As soon as I add an extension kubectl can't merge kubeconfigs anymore. Asked in #sig-api-machinry and got no helpful response.

So this added the version to the cluster name name@version. Should be totaly invisible to the user and we can use this locally to switch to the correct kubectl version.

One caveat: As clusters get upgraded we will leave old cluster definitions dangeling in the kubeconfig but yolo.


databus23:
A shortened kubeconfig returned by kubernikus then would look like this:
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: BlaBlaBla...
    server: https://mycluster-api.example.com
  name: mycluster@1.15.2
contexts:
- context:
    cluster: mycluster@1.15.2
    user: D062284@nocloud-15
   name: mycluster
...
```

PR #464, opened 2019-08-23
Kubernetes dashboard for kubernikus clusters

onuryilmaz:
This PR adds Kubernetes Dashboard to kubernikus clusters.

### Approach:
- [Kubernetes dashboard](https://github.com/kubernetes/dashboard), [a proxy](https://github.com/keycloak/keycloak-gatekeeper) and a [dex](https://github.com/sapcc/dex/tree/ccloud) instance added to `kube-master` chart.
  - Kubernetes dashboard is a plain installation of the official repository.
  - Dex is configured to work with the keystone and authentication scope of kluster project.
  - Proxy is the entrypoint to Kubernetes dashboard which handles authentication via Dex and passes headers to dashboard.
  - A similar setup: https://itnext.io/protect-kubernetes-dashboard-with-openid-connect-104b9e75e39c 
- `OIDC` flags are added to `apiserver`: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server
- A client secret is generated within kluster secret to be used between clients (dashboard proxy, `kubectl` etc.) and dex. In addition, migration step is added.
- Cluster role bindings with Openstack roles (`kubernikus:dashboard-openstack-kubernetes-admin` and `kubernikus:dashboard-openstack-kubernetes-member`) are seeded into customer klusters. In addition, migration step is added.
- `kubernikus-dashboard` chart is added to be installed in `kubernikus` namespace. 
  - It has a wildcard ingress and a secret to be signed by `vice-president`. The secret will be used by all the dashboards and dex instances.
  - It has a `kubernikus-dashboard-keystone` secret. All `dex` instances are using to authenticate with keystone. 

### Memory Usage:
Memory usage of the additional 3 pods seem to be around 20M/each for a couple of installations:
dashboard
dex
dex

### Demo:
- There is a live cluster on `admin/kubernikus-onur` which is scoped into `eu-nl-1/monsoon3/cc-demo`:
https://dashboard-dev-3e0fd3f8e9ec449686ef26a16a284265.ingress.kubernikus.eu-nl-1.cloud.sap
Note: `CNAME` of the zone is manually changed for testing and demo.


onuryilmaz:
> It would be great if we could reuse Dex both for Keystone and LDAP auth. In case of the bare-metal virtual APIs, we need to configure an LDAP connector.
> 
> Can we make it that we can pass in the connector config dynamically?

I can update the `connector` part to fill from Helm values:
https://github.com/sapcc/kubernikus/pull/464/files#diff-d238b1eb95a4202a4d4d6b0ca52509dcR136

BugRoger:
> I can update the connector part to fill from Helm values
👍

BugRoger:
It's either or in this case. But I think, this is what the selection looks like, if there are multiple options:

image

ruvr:
Yes. Assuming that you have client-scenarios that will allow a manual browser interaction screen-flow back-and-forth and if so, the user is aware what authorisation he will get where.. Maybe expecting too much of the user?

onuryilmaz:
> > I can update the connector part to fill from Helm values
> 
> 👍

Related `configmap` is updated to configure `keystone` and `ldap`:
https://github.com/sapcc/kubernikus/pull/464/files#diff-d238b1eb95a4202a4d4d6b0ca52509dcR163

`bindPW` is handled similar to `keystone` admin password:
https://github.com/sapcc/kubernikus/pull/464/files#diff-0475aa5b4c4cc95d3b803d6b182385a2R1

/cc @databus23

BugRoger:
Nice, thanks. Looks good. I'm wondering if we should also add a static password options, for the case that all else fails. Like so:

https://github.com/dexidp/dex/blob/master/examples/config-dev.yaml#L90-L95

pkg/controller/ground/bootstrap.go
```diff
@@ -310,3 +314,45 @@ func SeedAutoRenewalNodeCertificates(client clientset.Interface) error {
 		},
 	})
 }
+
+func SeedDashboardOpenstackClusterRoleBindings(client clientset.Interface) error {
+
+	err := bootstrap.CreateOrUpdateClusterRoleBinding(client, &rbac.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "kubernikus:dashboard-openstack-kubernetes-admin",
```
databus23: This role is not really dashboard specific, no? This role will also apply for users Kubectl that authenticate against dex afais. So maybe this should be called differently? Like `openstack-kubernetes-admin`. Also the function maybe should be called. `SeedOpenStackClusterRoleBindings` os something.

onuryilmaz: 👍 Fixed

onuryilmaz:
> Nice, thanks. Looks good. I'm wondering if we should also add a static password options, for the case that all else fails. Like so:
> 
> https://github.com/dexidp/dex/blob/master/examples/config-dev.yaml#L90-L95

I have enabled the `staticPasswords`. 
The flow works from **browser** for both `Login with Email` and `Login with SAP Converged Cloud`:
Screen Shot 2019-08-29 at 23 39 21

However, as @ruvr  mentioned, from CLI, only the connector configured as `passwordConnector` at `oauth2` section works:
https://github.com/sapcc/kubernikus/pull/464/files#diff-d238b1eb95a4202a4d4d6b0ca52509dcR200
- Currently, user for `staticPassword` is not working from non-interactive flow:
`kubectl-logon -u kubernikus@cloud.sap -p $PASSWORD` 
- Working:
`kubectl-logon -u $OS_USERNAME -p $OS_PASSWORD`

*I have manually changed the `passwordConnector` to `local` and tested for vice versa*

BugRoger:
That means, in case of LDAP/Keystone failure, we can login into the Dashboard but not the API?

ruvr:
Which is what I was trying to indicate with my previous comments. Logging in is more than just validation credentials. Depending on the connector/context, you'll get different permissions (or no permissions in case of static-password)

onuryilmaz:
We can authenticate against API, with a little hack, I have added the specific user in static password to RBAC:
https://github.com/sapcc/kubernikus/pull/464/files#diff-cc1a3a24eac74d6e270f5b17afead1c4R335

For `oauth2`, there is only one `passwordConnector` defined here:
https://github.com/sapcc/kubernikus/pull/464/files#diff-d238b1eb95a4202a4d4d6b0ca52509dcR200
We can login via CLI by passing username and password to only the `connector` mentioned above.

onuryilmaz:
With the latest commits:
## Dashboard flag
- `spec.dashboard` flag added to explicitly enable Dashboard & Dex related components.
- Controllers are updated to handle changes on the flag value. For instance, the following scenario are tested manually:
  - Create a cluster -> Wait until it is `Running`-> Update `spec.dashboard` to `true` -> Dashboard is created by updating the chart and URL is added to status -> Update `spec.dashboard` to `false` -> Dashboard and related parts are disabled by updating the chart -> Dashboard URL is removed in status


## API Endpoint for OIDC:
- A new endpoint added under `api/v1/clusters/$CLUSTER/credentials/oidc`
- The response `kubeconfig` is ready to be used by `kubectl logon` plugin at `monsoon/kubectl-logon`
  - It has no user specific information, since `token` and `refresh_token` will be added by `kubectl` plugin.

charts/kube-master/templates/dex.yaml
```diff
@@ -0,0 +1,264 @@
...
+      app: {{ include "master.fullname" . }}-dex 
+      release: {{ .Release.Name }}  
+  template:
+    metadata:
+      labels:
+        app: {{ include "master.fullname" . }}-dex 
+        release: {{ .Release.Name }} 
+    spec:
+      initContainers:
+      - image: xmartlabs/htpasswd
```
databus23: Extra points for going the extra mile to avoid leaking the secret in the ConfigMap. ❤️ 

But I think this is an awful lot of yaml and a basically untrusted container we add here.
So let me provide an alternative suggestion without an init container:
* I would generate the bcrypt hash in the operator when the helm values are computed. This frees us from  the `htpasswd` dependency during runtime
* Then I see three options where to put the password:
1. but the whole config in a secret instead of a ConfigMap and mount that
2. leave it in a ConfigMap and do the SED magic in the main container before starting dex
3. but the hashed password in the configmap and rely on bcrypt being reasonably safe from decryption

All three options are fine by me and I guess the third one would be the simplest with an arguably negligible decrease in security.

databus23:
Thanks a lot for the effort! This looks very good already. I added couple of remarks for discussion.

charts/kube-master/templates/dashboard.yaml
```diff
@@ -0,0 +1,147 @@
+{{ if and .Values.dashboard .Values.dashboard.enabled }}
```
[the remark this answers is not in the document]
onuryilmaz: 👍 , I will fix it!

charts/kube-master/templates/dex.yaml
```diff
@@ -0,0 +1,264 @@
+{{ if and .Values.dashboard .Values.dashboard.enabled }}
```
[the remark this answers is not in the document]
onuryilmaz: Then the flag options will be like:
- **dex** (true/false) 
- if **dex**  is true
  - **dashboard and dashboard proxy** (true/false)

I guess it will add another field in `KlusterSpec` or do you suggest any other option?

charts/kube-master/templates/dex.yaml
onuryilmaz: In the third option, if you are suggesting the following, I can implement and test it: 
- Hashing the `DashboardStaticPassword` from the secret in `KlusterToHelmValues` and passing the hashed value as to the `kube-master` chart.

Should we also pass `DashboardClientSecret` as raw Helm value? It is also replaced by `sed` in the init container and there is no hashing for it.

onuryilmaz: It is implemented with https://github.com/sapcc/kubernikus/pull/464/commits/8bd00d13b31c83eddd3a01424bff555b14465b38

charts/kube-master/templates/dex.yaml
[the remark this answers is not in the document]
onuryilmaz: It is implemented by: https://github.com/sapcc/kubernikus/pull/464/commits/1a4761706b52e0e6c7d145206b911197dc7ad45a

Manually tested configurations:
Dex: true
- Dashboard: false  --> Only dex is installed, CLI access with OIDC login works.
- Dashboard: true --> Dashboard and proxy is additionally installed.

databus23:
@onuryilmaz Something went wrong with the last commit, there is now a lot more changes related to the webpage that have nothing todo with this PR.

charts/kube-master/templates/api.yaml
```diff
@@ -176,6 +176,13 @@ spec:
             - --tls-cert-file=/etc/kubernetes/certs/tls-apiserver.pem
             - --tls-private-key-file=/etc/kubernetes/certs/tls-apiserver-key.pem
             # --tls-sni-cert-key=/etc/kubernetes/certs/tls-sni.pem,/etc/kubernetes/certs/tls-sni.key
+            {{ if.Values.dex.enabled }}
```
BugRoger: typo? Missing space after if.

onuryilmaz: fixed

charts/kube-master/values.yaml
```diff
@@ -80,3 +80,51 @@ scheduler:
       memory: 512Mi
 
 revisionHistoryLimit: 3
+
+dex:
+  enabled: false
+  image:
```
BugRoger: Please move the image values to to the `images.yaml` https://github.com/sapcc/kubernikus/blob/master/charts/images.yaml#L2-L9

To properly reference it again we need this (arguably ugly) helper methods to: 
https://github.com/sapcc/kubernikus/blob/master/charts/kube-master/templates/_helpers.tpl#L31-L38

The advantage of this is that we can easily update the image versions at a central place. Also it's possible to configure different versions for different Kubernetes releases. Could be useful for the Dashboard with `v1.10.1` as default and `v2.0.0-beta4` for pre-release version.

charts/kube-master/values.yaml
```diff
@@ -80,3 +80,51 @@ scheduler:
...
+  dns:
+    zone: ingress.kubernikus
+    domain: cloud.sap
+  ingressSecret: kubernikus-dex
+
+dashboard:
+  enabled: false
+  
+  image:
+    repository: kubernetesui/dashboard
```
BugRoger: Same here as above

charts/kube-master/values.yaml
```diff
@@ -80,3 +80,51 @@ scheduler:
...
+    tag: v2.0.0-beta3
+  proxy:
+    repository: quay.io/keycloak/keycloak-gatekeeper
```
BugRoger: same here as above

charts/kube-master/templates/dex.yaml
```diff
@@ -0,0 +1,246 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ include "master.fullname" . }}-dex 
+      release: {{ .Release.Name }}  
+  template:
+    metadata:
+      labels:
+        app: {{ include "master.fullname" . }}-dex 
+        release: {{ .Release.Name }}
```
BugRoger: Please add a checksum annotation here. In case the configmap changes this will then trigger a restart. See for example: https://github.com/sapcc/kubernikus/blob/master/charts/kube-master/templates/_helpers.tpl#L31-L38

onuryilmaz: dex configmap is separated and checksum is added to the deployment: https://github.com/sapcc/kubernikus/pull/464/commits/ab997d3abb599ada9b7e84e652f4b8b80111e944#diff-d238b1eb95a4202a4d4d6b0ca52509dcR22

charts/kube-master/values.yaml
onuryilmaz: fixed by: https://github.com/sapcc/kubernikus/pull/464/commits/ab997d3abb599ada9b7e84e652f4b8b80111e944#diff-d238b1eb95a4202a4d4d6b0ca52509dc

charts/kube-master/values.yaml
onuryilmaz: fixed by: https://github.com/sapcc/kubernikus/pull/464/commits/ab997d3abb599ada9b7e84e652f4b8b80111e944#diff-d238b1eb95a4202a4d4d6b0ca52509dc

charts/kube-master/values.yaml
onuryilmaz: fixed by: https://github.com/sapcc/kubernikus/pull/464/commits/ab997d3abb599ada9b7e84e652f4b8b80111e944#diff-d238b1eb95a4202a4d4d6b0ca52509dc

onuryilmaz:
> Found a few more nitpicks. Sorry this is dragging so long. Let's get this boat moving again.

I have tested starting with 1.11.9 (default) and incrementing the cluster version and checking the dashboards.

I have updated the images using the compatibility matrices from https://github.com/kubernetes/dashboard/releases:
```
1.11.9: v1.10.1
1.12.10: v2.0.0-beta1
1.13.9: v2.0.0-beta1
1.14.5: v2.0.0-beta1
1.15.2: v2.0.0-beta4
```

PR #465, opened 2019-08-28
Exclude unavailable AVZs from default

jknipper:
This PR excludes unavailable AVZs from default selection. Apparently there will be a new zone `na-us-1d` that is unavailable right now. Currently Kubernikus tries to deploy in that AVZ when there is none choosen.

### DIFF
--- a/pkg/client/openstack/project/client.go
+++ b/pkg/client/openstack/project/client.go
@@ -233,7 +233,9 @@ func (c *projectClient) getAvailabilityZones() ([]models.AvailabilityZone, error
 		return nil, err
 	}
 	for _, zone := range zones {
-		result = append(result, models.AvailabilityZone{Name: zone.ZoneName})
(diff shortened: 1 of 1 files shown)


databus23:
@jknipper The test are failing for this PR, probably the tests need updating?

pkg/api/models/availability_zone.go
```diff
@@ -17,6 +17,9 @@ type AvailabilityZone struct {
 
 	// name
 	Name string `json:"name,omitempty"`
+
+	// state
+	State bool `json:"state,omitempty"`
```
databus23: What does `State: true` mean? We should either have this a string field carrying the state name or change this field to `Available` or `Active` of state bool.

jknipper:
@databus23 I changed this completly, I'm now checking for availability in `getAvailabilityZones()`.

PR #466, opened 2019-08-29
Add prometheus target to etcd

jknipper:
This PR enables scraping of etcd-backup-restore metrics. @auhlig Will scraping still work for our customer klusters in k-region with this annotation?

### DIFF
--- a/charts/kube-master/charts/etcd/templates/deployment.yaml
+++ b/charts/kube-master/charts/etcd/templates/deployment.yaml
@@ -69,7 +69,11 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8080"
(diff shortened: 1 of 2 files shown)


charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -69,6 +69,7 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8080"
+        prometheus.io/targets: "kubernetes"
```
auhlig: It's matching the tier you know from slack, so `kubernikus` in this case

jknipper:
@auhlig Like this?

jknipper:
This is not needed anymore.

PR #470, opened 2019-09-03
Add etcd backup to S3

jknipper:
This PR adds capability to store etcd backup to S3.


jknipper:
@databus23 I changed the default behavior of kluster spec is not set. Please check.

PR #477, opened 2019-09-27
Remove projectID from cluster spec

databus23:
The projectID should be fixed to the project the cluster was created in, we track it in the account label of the `Kluster` object.


kayrus:
@databus23 what is an admin wants to create a cluster inside the specific tenant from the admin scope?

databus23:
Thats not a supported use case

databus23:
go away stale bot

BugRoger:
tumbleweed

PR #479, opened 2019-10-04
Add kluster annotation to specify addtional apiserver SANs

databus23:


pkg/controller/certs/certs.go
```diff
@@ -7,7 +7,7 @@ import (
 	kitlog "github.com/go-kit/kit/log"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
+	v1 "github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
```
BugRoger: Linter?

databus23: yup

PR #485, opened 2019-10-23
fix typo

majewsky:
Noticed this while reading some of your code for reference. Am I right in assuming that this should be `applyErr` instead of `err`?


databus23:
Thank @majewsky

PR #489, opened 2019-10-31
Adds an advertisePort option

SchwarzM:
To control the endpoint in default/kubernetes we need to control the port as well as the 
address.


pkg/client/kubernetes/shared_client_factory.go
```diff
@@ -80,13 +80,14 @@ func (f *sharedClientFactory) ClientFor(k *kubernikus_v1.Kluster) (clientset kub
 	// We need to provide a custom dialer to add the kluster namespace to the dns resolution because the
 	// apiserver cert is missing an SAN for $kluster.$namespace
 	if os.Getenv("KUBERNETES_SERVICE_HOST") != "" {
-		apiHost = fmt.Sprintf("https://%s:6443", k.Name)
+		port := os.Getenv("KUBERNETES_SERVICE_PORT")
```
databus23: I think this is not correct. you will get the port of the parent cluster where the operator is running in. this will be the same port every time, for every kluster running in this kubernikus.

BugRoger:
Remove WIP to make bot happy. Squash and rebase.

PR #492, opened 2019-11-11
Allow to specify empty clusterCIDR

databus23:
If a cluster is created with an explitly empty nodeCIDR "" the node CIDR allocation will be disabled.

If its not set it will still be populated with the default clusterCIDR range.


SchwarzM:
https://github.com/sapcc/kubernikus/blob/master/pkg/templates/node_1.14.go#L338
can never be hit with clusterCIDR nil ?

pkg/apis/kubernikus/factory.go
```diff
@@ -53,8 +53,14 @@ func (klusterFactory) KlusterFor(spec models.KlusterSpec) (*v1.Kluster, error) {
 		},
 	}
 
-	if k.Spec.ClusterCIDR == "" {
-		k.Spec.ClusterCIDR = DEFAULT_CLUSTER_CIDR
+	//If the cluster CIDR was not set at all we default it,
+	//if was set to empty string explicitly we null it
```
SchwarzM: this is weird. i have no clue if this holds up in the templates
i think most / all clusters with empty CIDR have no nodes but i am unsure if we actually prevent it.

pkg/api/models/kluster_print.go
```diff
@@ -31,7 +31,10 @@ func (k Kluster) Print(format printers.PrintFormat, options printers.PrintOption
 func (k Kluster) printHuman(options printers.PrintOptions) {
 	fmt.Println("Cluster name: ", k.Name)
 	fmt.Println("Cluster state: ", k.Status.Phase)
-	fmt.Println("Cluster CIDR: ", k.Spec.ClusterCIDR)
+	if k.Spec.ClusterCIDR != nil && *k.Spec.ClusterCIDR != "" {
+		fmt.Println("Cluster CIDR: ", *k.Spec.ClusterCIDR)
+	}
+
```
SchwarzM: If i read this correct the clusterCIDR will just be missing from the output.
Spotting a missing line in debug is way harder than printing: *** UNSET *** or something similar.

databus23: Ok I think I've addressed this. I now print the cluster CIDR if its not `nil`. Which should not happen at all because. I now also keep it as an empty string in storage (CRD). This means the only time this should be actually nil is during cluster creation.

pkg/apis/kubernikus/factory.go
databus23: I made an additional change that you can only set an empty clusterCIDR if `noCloud: true`, so its unlikely that this should happened by accident.

databus23:
@SchwarzM I made a few changes based on your feedback.
We now keep the empty string for `clusterCIDR` in storage. And setting it to empty is only allowed for `noCloud` clusters.

PR #493, opened 2019-11-14
allows for dash and dex paramaters to be disabled

BugRoger:


pkg/api/handlers/get_cluster_credentials_oidc.go
```diff
@@ -33,7 +33,7 @@ func (d *getClusterCredentialsOIDC) Handle(params operations.GetClusterCredentia
 		return NewErrorResponse(&operations.GetClusterCredentialsDefault{}, 500, err.Error())
 	}
 
-	if !kluster.Spec.Dex {
+	if kluster.Spec.Dex == nil || !*kluster.Spec.Dex {
```
databus23: using the `BoolValue(v *bool) bool` function from the swag package could help clean up all these if statements
```suggestion
	if ! swag.BoolValue(kluster.Spec.Dex) {
```

BugRoger: Thanks. Nice tip!

PR #494, opened 2019-11-18
Fix cert rotation on node startup

jknipper:
This PR fixes certificate rotation on nodes <1.11 on startup.


jknipper:
I added restart of fix on failure, also 1.8/1.9 node templates are included. @databus23 Please have a look!

pkg/templates/node_1.10.go
```diff
@@ -212,6 +212,20 @@ systemd:
...
+      command: start
+      enable: true
+      contents: |
+        [Unit]
+        Description=Fix kubelet certificate rotation
+        After=kubelet.service
+        [Service]
+        Type=simple
+        Restart=on-failure
+        RestartSec=5
```
databus23: Maybe we should relax this to 60 or even 300 seconds? There is not need to retry that often, we have no hurry in fixing it, no? If this fails constantly, retrying every 5 seconds seems excessive

jknipper: You are right, there is no hurry. I changed this to 300s.

PR #495, opened 2019-11-19
allow setting annotations on node(pool)

auhlig:


pkg/templates/ignition.go
```diff
@@ -98,6 +101,9 @@ func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, pool *models.Node
 		for _, userLabel := range pool.Labels {
 			nodeLabels = append(nodeLabels, userLabel)
 		}
+		for _, userAnnotation := range pool.Annotations {
+			nodeAnnotations = append(nodeAnnotations, userAnnotation)
```
databus23: I don't understand where this is used. This is a new variable and it I'm missing how this is passed to ignition.

databus23:
Can you elaborate why this is needed a bit? We already label each node pool with something. it seems to me this just adds another form of tag to a node pool. Couldn't you also use the nod pool label for selecting what do do?

auhlig:
Want to control k8s operators on node level, like `give-me-a-fip: "true"` annotation the operator would do so.
Would also work with `labels` of course. I though `annotations` would make since the above is non-identifying to node.

auhlig:
scratch that. not possible to register node with annotations 😓 just using labels now

PR #497, opened 2019-11-28
Create extra values field in kluster secret

SchwarzM:
We need a way to customize the helm-chart values without exposing everything into the API.

This adds a field into the kluster secret to hold a map of extra values which get deep merged into
the helm values used to render to helm.

This lets us customize resources and additional flags on the apiserver, controller-manager, scheduler etc. without exposing them via the API.

### DIFF
--- a/pkg/apis/kubernikus/v1/secret.go
+++ b/pkg/apis/kubernikus/v1/secret.go
@@ -18,7 +18,7 @@ type Secret struct {
 
 	Certificates
 
-	ExtraValues map[interface{}]interface{} `json:"extra-values,omitempty"`
+	ExtraValues string `json:"extra-values,omitempty"`
 }
 
 func NewSecret(secret *corev1.Secret) (*Secret, error) {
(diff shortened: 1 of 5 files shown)


pkg/apis/kubernikus/v1/secret.go
```diff
@@ -17,6 +17,8 @@ type Secret struct {
 	DexStaticPassword string `json:"dex-static-password,omitempty"`
 
 	Certificates
+
+	ExtraValues map[interface{}]interface{} `json:"extra-values,omitempty"`
```
databus23: I'm not sure this will fly when reading this from the underlying Kubernetes secret.
The problem is secrets don't allow nested son structures. the field `extra-values` has to be of type `string` or `[]byte`. I think the serialisation, deserialisation will bomb for this struct.

I think what we need to do is make this a `string` field and `yaml.Unmarshal` it in `KlusterToHelmValues`.

Maybe we could also write a test that uses a kubernetes fake client to make sure this works.
You can probably add it the pkg/apis/kubernikus/v1/secret_test.go

SchwarzM: you are correct.

corrected.

databus23:
Ok this looks good but I think what is still missing is actually calling MergeMaps in `KlusterToHelmValues` no?

PR #508, opened 2020-01-27
Add CSIInlineVolume support

jknipper:
This PR enables CSI inline volume support in apiserver and kubelet for Kubernetes >=v1.15. See https://kubernetes.io/blog/2020/01/21/csi-ephemeral-inline-volumes/ for more details.

### DIFF
--- a/pkg/templates/node_1.14.go
+++ b/pkg/templates/node_1.14.go
@@ -456,6 +456,7 @@ storage:
           rotateCertificates: true
           featureGates:
             NodeLease: false
+            CSIInlineVolume: true
     - path: /etc/kubernetes/kube-proxy/config
       filesystem: root
       mode: 0644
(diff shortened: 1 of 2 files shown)


pkg/templates/node_1.14.go
```diff
@@ -456,6 +456,7 @@ storage:
           rotateCertificates: true
           featureGates:
             NodeLease: false
+            CSIInlineVolume: true
```
BugRoger: Is this the only change required on the kublet to make CSI work?

jknipper: This depends a lot on the used CSI driver. At least https://github.com/kubernetes-csi/csi-driver-image-populator worked with these settings.

BugRoger:
We don't need this anymore with 1.16, do we?

jknipper:
> We don't need this anymore with 1.16, do we?

@BugRoger No, this is enabled by default in 1.16.

PR #510, opened 2020-01-29
email suffix for technical users

onuryilmaz:
Background:
Technical users have no email addresses and they cannot login to dex due to the missing attributes: https://github.com/sapcc/dex/blob/master/connector/ldap/ldap.go#L344

With this PR, three changes are done:
- email suffix (`cloud.sap`) is added as a workaround, so that email addresses will be structured with `username@cloud.sap` for every users including technical users: https://github.com/sapcc/dex/blob/master/connector/ldap/ldap.go#L345
- `nameAttr` is switched to `cn`  (such as I/D/Technical User)
- `oidc-username-claim` flag switched to `name` (such as I/D/Technical User) so that they can be referenced in RBAC.

Follow-up tasks:
- Extend groupSearch in `cc/kube-secrets` to allow logging in users who are also not in `CP_CONV_K8S_USERS`.
- Allow adding additional users in `cc-rbac` chart inside values: https://github.com/sapcc/helm-charts/tree/master/system/cc-rbac

### DIFF
--- a/charts/kube-master/Chart.yaml
+++ b/charts/kube-master/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: kube-master
-version: 2.0.3
+version: 2.0.4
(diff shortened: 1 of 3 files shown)


onuryilmaz:
Changes are manually tested in `v-qa-de-1` and `qa-de-1` as they rely on LDAP login.

PR #517, opened 2020-02-14
Add flatcar support

jknipper:
This PR adds Flatcar Linux support:
- reboot outdated nodes
- replace nodes if image has changed in nodepool
- ~~use Flatcar image by default~~
- fixes e2e test

### DIFF
--- a/pkg/controller/servicing/lister.go
+++ b/pkg/controller/servicing/lister.go
@@ -10,6 +10,7 @@ import (
 	"github.com/sapcc/kubernikus/pkg/controller/config"
 	"github.com/sapcc/kubernikus/pkg/controller/nodeobservatory"
 	"github.com/sapcc/kubernikus/pkg/controller/servicing/coreos"
+	"github.com/sapcc/kubernikus/pkg/controller/servicing/flatcar"
 
 	"github.com/go-kit/kit/log"
 	core_v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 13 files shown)


pkg/controller/servicing/lister.go
```diff
@@ -194,14 +237,24 @@ func (d *NodeLister) Replace() []*core_v1.Node {
 
 	klusterVersion, err := version.ParseSemantic(d.Kluster.Status.ApiserverVersion)
 	if err != nil {
-		d.Logger.Log(
-			"msg", "Couldn't parse Kluster version. Skipping node upgrade.",
-			"err", err,
-		)
-		return found
+		klusterVersion = nil
```
BugRoger, line 240: I'm confused about this change. When would this make a difference? If the `klusterVersion` is nil it will skip all nodes. Or am I missing something?

pkg/controller/servicing/lister.go
```diff
@@ -194,14 +237,24 @@ func (d *NodeLister) Replace() []*core_v1.Node {
...
 	}
 
 	for _, node := range upgradable {
+		if util.IsCoreOSNode(node) && util.IsFlatcarNodePool(nodeNameToPool[node.GetName()]) {
```
BugRoger, line 244: This confused me also. I suppose, the idea is that we somehow flip the `image` in a nodePool from CoreOS to Flatcar and this marks the node as `upgradable`?

swagger.yml
```diff
@@ -515,7 +515,7 @@ definitions:
       image:
         x-nullable: false
         type: string
-        default: coreos-stable-amd64
+        default: flatcar-stable-amd64
```
BugRoger, line 518: This change effectively makes all new clusters use flatcar. Can we extract it in an extra PR for visiblity?

pkg/controller/servicing/lister.go
jknipper: I doesn't make a real difference, currently nodes get replaced with flatcar even if `klusterVersion` can't be determined.

pkg/controller/servicing/lister.go
jknipper: If current node image differs from nodepool it will be marked as upgradable (and eventually replaced).

BugRoger:
This is going in now.

PR #519, opened 2020-02-18
Make sure all new clusters have ValidateProxyRedirects enabled

databus23:
This requires a minimum cluster version of 1.12.
The apiserver refuses to create clusters with versions that are not "supported".

In this PR we also bump the default version to 1.15.9


databus23:
Approved verbally by @BugRoger

PR #524, opened 2020-04-29
Use flatcar by default

jknipper:
From now on flatcar will be used as default node image. A kluster upgrade will switch nodepools from coreos to flatcar.


jknipper:
@databus23 Yeah, already on it.

PR #525, opened 2020-04-30
Add apiserver flags for TokenRequestProjection and TokenRequest API

mvladev:
This enables [token volume projection to work][1] on clusters `>= 1.12`.

`kubernetes` is used as a a shared audience between different clusters as it allows for the same `TokenRequest` to be used on each one of them:

```yaml
apiVersion: authentication.k8s.io/v1
kind: TokenRequest
spec:
  audiences:
  - my-custom-audience
  - kubernetes
  expirationSeconds: 600
```

This would return a token valid for both `kube-apiserver` and `my-custom-audience`. This can be used by some other third-party component to verify the the validity of the token with 

```yaml
apiVersion: authentication.k8s.io/v1
kind: TokenReview
spec:
  token: abcd.....
  audiences: 
  - my-custom-audience
```

Another popular component using this feature is [istio][2].

[1]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection
[2]: https://istio.io/blog/2019/trustworthy-jwt-sds/


mvladev:
ping @databus23 @BugRoger

BugRoger:
Thanks for the PR, Martin. 

I suppose, in your example it is required to have `my-custom-audience` configured on the APIserver flags too? Do you have experience about the most common audiences? I think `istio` is commonly used as well. 

We don't really have a mechanism to configure this per user.

mvladev:
> I suppose, in your example it is required to have `my-custom-audience` configured on the APIserver flags too? 

No, it's not required to have this flag as well. When the token is received by the `kube-apiserver` it must have  at least `kubernetes` audience otherwise it's not a valid for it. Any other audiences are not relevant and are ignored in the context of the `kube-apiserver`.

Let me clarify about this - the main use-case is an application running in a K8S cluster which has received a token via the `TokenRequest` API. Now using this token my application can make calls to:

- the `kube-apiserver` as it's allowing the audience `kubernetes`.
- and `my-other-application` which runs somewhere else and it allows token with `my-custom-audience` audience. `my-custom-audience` is only known to my custom application and it's ignored by `kube-apiserver`.

mvladev:
In the [kube-apiserver](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) documentation:

> --api-audiences stringSlice Identifiers  of the API. The service account token authenticator will validate that  tokens used against the API are bound to at least one of these  audiences. If the --service-account-issuer flag is configured and this  flag is not, this field defaults to a single element list containing the  issuer URL.

I can also add another entry with the `--api-audiences` issuer URL as well.

PR #448, opened 2019-08-01
Update kubernetes dependencies to 1.12

databus23:

PR #528, opened 2020-05-20
Add support for 1.17/1.18/1.19

jknipper:
This PR adds support for k8s 1.17/1.18/1.19 to Kubernikus. It depends on #448 because node leases are now used.


charts/kube-master/templates/api.yaml
```diff
@@ -189,7 +191,9 @@ spec:
             - --enable-admission-plugins=PersistentVolumeLabel
 {{- end }}
 {{- if (semverCompare ">= 1.14" .Values.version.kubernetes) }}
+{{- if (semverCompare "< 1.17" .Values.version.kubernetes) }}
```
databus23, line 194: Interesting, why is this necessary?
This might conflict with hammertime which depends on frequent node updates to not kick in. 
In general we should switch to NodeLeases anyway but this requires updating the client-go in kubernetes.

jknipper: Apparently this feature needs to be turned on:
```
error: invalid argument "NodeLease=false" for "--feature-gates" flag: cannot set feature gate NodeLease to false, feature is locked to true
```

BugRoger:
Ressurect

jknipper:
I'm going to update this to the latest 1.17 release.

databus23 (approved):
Looks good. Great work!

PR #533, opened 2020-07-02
add support to scrape k8s metrics

defo89:
Prepare Kubernetes API components to be scraped with `ServiceMonitor` of Prometheus Operator in `kube-monitoring-virtual`.

Hence, metrics ports are exposed with additional services for metrics.

Additional labels were added because I didn't want to edit existing `app` labels.

### DIFF
--- a/charts/kube-master/charts/etcd/templates/deployment.yaml
+++ b/charts/kube-master/charts/etcd/templates/deployment.yaml
@@ -127,6 +127,9 @@ spec:
       containers:
         - name: etcd
           ports:
+          - containerPort: 2379
+            name: etcd
+            protocol: TCP
           - containerPort: 8081
             name: metrics
             protocol: TCP
--- a/charts/kube-master/charts/etcd/templates/deployment.yaml
+++ b/charts/kube-master/charts/etcd/templates/deployment.yaml
@@ -49,7 +49,8 @@ data:
       --client-cert-auth=true \
       --trusted-ca-file=/etc/kubernetes/certs/etcd-clients-ca.pem \
       --advertise-client-urls=https://${ETCD_IP}:2379 \
-      --listen-client-urls=https://0.0.0.0:2379
+      --listen-client-urls=https://0.0.0.0:2379 \
+      --listen-metrics-urls=http://0.0.0.0:8081
 {{- else }}
       --advertise-client-urls=http://${ETCD_IP}:2379 \
       --listen-client-urls=http://0.0.0.0:2379 \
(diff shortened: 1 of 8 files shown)


charts/kube-master/templates/controller-manager.yaml
```diff
@@ -82,6 +83,12 @@ spec:
               readOnly: true
       containers:
         - name: controller-manager
+{{- if semverCompare ">=1.12-0" .Capabilities.KubeVersion.GitVersion }}
```
databus23, line 86: can we change this to `">= 1.12"`? Thats what we are using in this deployment in a different location?

charts/kube-master/templates/scheduler.yaml
```diff
@@ -45,6 +46,12 @@ spec:
             name: {{ include "master.fullname" . }}
       containers:
         - name: scheduler
+{{- if semverCompare ">=1.12-0" .Capabilities.KubeVersion.GitVersion }}
```
databus23, line 49: same here I would suggest to change it to `">= 1.12"
`

charts/kube-master/templates/service-metrics.yaml
```diff
@@ -0,0 +1,46 @@
...
+  name: etcd-metrics
+spec:
+  ports:
+  - name: metrics
+    port: 8081
+    protocol: TCP
+    targetPort: 8081
+  selector:
+    component: etcd
+{{- end }}
```
databus23, line 46: I think the `{{ end }}` should move up to only exclude the controller-manager and scheduler for <1.12 as we enable metrics for etcd in all cases.

charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -52,7 +52,8 @@ data:
       --listen-client-urls=https://0.0.0.0:2379
 {{- else }}
       --advertise-client-urls=http://${ETCD_IP}:2379 \
-      --listen-client-urls=http://0.0.0.0:2379
+      --listen-client-urls=http://0.0.0.0:2379 \
+      --listen-metrics-urls=http://0.0.0.0:8081
```
databus23, line 56: hmm we have secure enabled by default now. so I think if we leave it like this we don't get metrics for etcd, no? I think we should add the metrics flag in all cases (outside of the if statement). What do you think?

defo89: you are right, I have missed the `.Values.secure.enabled` bit

PR #540, opened 2020-09-01
Switch ingress resources to cert-manager

jknipper:
This PR switches Kubernikus API and Dex ingress to use cert-manager for certificate issuance.


BugRoger (approved):
LGTM.

PR #542, opened 2020-09-23
Waits for Load Balancer Deletions via Finalizers

BugRoger:
This commit changes the wait behaviour when deleteing load balancers.
Previously service deleteion didn't use finalizers and no feedback was
given. The service disapeared immediately. The deorbiter used a fixed
timeout of 2min to wait for load balancer cleanup.

Since the loadbalancer driver in CCEE was changed to Octavia, that grace
period is not enough anymore. Deleteion takes longer and in case of 1.16
never completes if the LoadBalancer is not completely provisioned, e.g.
due to quota limitations. Especially a FIP is expected to be attached
and causes the deletion to never complete.

This fix depends on Kubernetes 1.17 to be fully functional. For
Kubernetes < 1.17 the Deorbiter will wait 24h for the service to be
deleted. For 1.16 it is possible to set the `use-octavia` flag as
a workaround.

### DIFF
--- a/pkg/controller/deorbit/deorbiter.go
+++ b/pkg/controller/deorbit/deorbiter.go
@@ -17,6 +17,7 @@ import (
 	v1 "github.com/sapcc/kubernikus/pkg/apis/kubernikus/v1"
 	"github.com/sapcc/kubernikus/pkg/controller/config"
 	"github.com/sapcc/kubernikus/pkg/controller/metrics"
+	"github.com/sapcc/kubernikus/pkg/util"
 )
 
 type SelfDestructReason string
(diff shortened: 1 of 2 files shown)


databus23:
@BugRoger Whats this this little nugget I found in the PR list? Is it ready to be merged?

BugRoger:
Can go in, I guess.

databus23:
Should we maybe keep the current behaviour for <1.16 clusters (we do set use-octavia) for a while now. I just would like to avoid customer inquiries about the old cluster stuck in terminating.

databus23:
rebased on master

databus23 (approved):

PR #543, opened 2020-09-29
move dex and keycloak-gatekeeper to keppel

defo89:


databus23 (approved):

databus23 (changes requested):
the dev image should have an updated tag that reflects the commit which includes the change to the docker file

defo89:
> the dex image should have an updated tag that reflects the commit which includes the change to the docker file

That makes sense, updated the image tag.

PR #546, opened 2020-10-21
turns Role into ClusterRole

BugRoger:
There is no `Role` with the name `edit`. By default, users which are assigned the `kubernikus:member` RoleBinding, will get that the `edit` Role can't be found. 

This was either always broken. Or the `Role` was changed to a `ClusterRole` without us noticing. It is a user-facing role and comes with Kubernetes itself. For more info, see here: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

Either way, it needs to be a `ClusterRole` now.


BugRoger:
Testing that change is not straight forward, as we are `kubernikus:admin` by default. I did test it by making the change, then removing the `kubernikus:admin` ClusterRoleBinding in a test cluster. Should be good to go.

BugRoger:
@jknipper any concerns merging this? I think, it just came up in #kubernikus-users again. Twice.

jknipper (approved):
No, I will merge this.

PR #549, opened 2020-11-04
charts/*: avoid deprecated API versions

majewsky:
As it is now, several objects would only transition to the new API versions in the Helm release once their hosting clusters update to k8s 1.16. However, Helm only supports transitioning from deprecated API versions *before* upgrading to a k8s that removes the API versions in question:

> The best practice is to upgrade releases using deprecated API versions to supported API versions, prior to upgrading to a kubernetes cluster that removes those API versions. [(Source)][1]

This patch is the very first step towards ensuring this. A separate question is how to force a `helm upgrade` onto all existing Helm releases of kubernikus, kubernikus-dex and especially kube-master.

[1]: https://helm.sh/docs/topics/kubernetes_apis/


majewsky:
Why do we even have the stale bot in the first place?

jknipper (approved):

PR #551, opened 2020-11-11
Change container bridge mtu

SchwarzM:
according to: https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/
the kubenet plugin assigns the lowest mtu on any active interface of the node to the container bridge.

in our case it is 1500 on the kubernikus interface.

we can use network-plugin-mtu option to the kubelet to configure it manually ( in this pullreq ) or reconfigure kubernikus
and let the autodiscover happen.

p.s. tested this on a manual change cluster


jknipper:
There is a problem when this change is applied. Nodes can't register because they can't reach the Apiserver.

SchwarzM:
this just changes a kubelet commandline ... nothing to do with networking of the node ... 🤷

jknipper:
It tested again and now the nodes are registering. Seems we had some network glitch somewhere else.

jknipper (approved):

Issue #529, opened 2020-05-20
dex: project domain name with different scoped tokens


onuryilmaz:
When the user is scoped to another project while creating the kluster, the domain name here is set wrongly:
https://github.com/sapcc/kubernikus/commit/aa60af8f638783f95f7090f30e72a6b5a2270b94#diff-2cedbfff1a4408572cc39c33d371b00dR546

For instance, if the user token is scoped to `Default` and created kluster in domain X, code above sets the project domain name as `Default`.

PR #553, opened 2020-11-13
Ensure kluster domain is project domain

onuryilmaz:
Fixes https://github.com/sapcc/kubernikus/issues/529


databus23:
Thanks a lot.
Can you also remove the "domain" label here?
https://github.com/sapcc/kubernikus/blob/master/pkg/api/handlers/create_cluster.go#L141

Maybe change the `creator` annotation instead to `fmt.Sprintf("%s/%s", principal.Name, principal.Domain)`

onuryilmaz:
@databus23 changes pushed

databus23 (approved):

databus23 (approved):

PR #557, opened 2020-12-01
[WIP] Detect cert rotation in wormhole

jknipper:


jknipper:
Will be handled in #566

PR #561, opened 2020-12-18
move etcd images to images.yaml, bump to 3.3.14

databus23:
I removed 1.10.8 from the images.yaml because we have no clusters with that version anymore.

### DIFF
--- a/charts/kube-master/charts/etcd/values.yaml
+++ b/charts/kube-master/charts/etcd/values.yaml
@@ -3,10 +3,10 @@
 # Declare variables to be passed into your templates.
 images:
   etcd:
-    repoitory: sapcc/etcd
-    tag: v3.3.13
+    repository: sapcc/etcd
+    tag: v3.3.14
   etcdBackup:
-    repoitory: sapcc/etcdbrctl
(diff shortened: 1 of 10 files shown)


charts/kube-master/charts/etcd/values.yaml
```diff
@@ -1,10 +1,13 @@
 # Default values for etcd.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
-image:
-  repository: sapcc/etcd
-  tag: v3.3.13
-  pullPolicy: IfNotPresent
+images:
+  etcd:
+    repoitory: sapcc/etcd
```
jknipper, line 6: @databus23 This typo makes the chart test fail.

charts/kube-master/charts/etcd/values.yaml
```diff
@@ -1,10 +1,13 @@
...
+    tag: v3.3.13
+  etcdBackup:
+    repoitory: sapcc/etcdbrctl
```
jknipper, line 9: Here as well.

jknipper (approved):
I tested the upgrade case, good lookin'!

PR #563, opened 2021-01-20
Add cinder CSI plugin support

jknipper:
This PR adds cinder CSI support for >=1.20, it depends on https://github.com/sapcc/kubernikus/pull/564
It also introduces resizing and snapshots of volumes.


jknipper:
Todo: volume cleanup on cluster deletion

PR #571, opened 2021-04-30
Bump k8s deps to 1.16.15, switch from glide to go modules

databus23:
Compiles, but I haven't tested it yet


jknipper:
I updated missing dependencies from master branch and bumped to go 1.16. @BugRoger Can you maybe also take a look? I would like to merge this.

contrib/all/Dockerfile.apiserver
```diff
@@ -1,6 +1,6 @@
 ARG IMAGE
 
-FROM golang:1.13 as builder
+FROM keppel.eu-de-1.cloud.sap/ccloud-dockerhub-mirror/library/golang:1.16-alpine as builder
```
majewsky, line 3: Two things: First, you want `1.16-alpine3.13`. The newest Alpine has problems with the Docker version that our Concourse uses: https://gitlab.alpinelinux.org/alpine/aports/-/issues/12519

Second, I'd recommend not hardcoding our internal Docker Hub mirror here. Leave the references to the original Docker Hub images here, and use `ci/shared/task-inject-dockerhub-mirror.yaml` in the pipeline (you can look at the `build` task in the Keppel pipeline for a usage example).

databus23: Thanks! I modified the image reference. The task injection seems like a hassle. I keep it in mind :)

PR #577, opened 2021-06-07
[WIP] Delete unschedulable nodes first on nodepool resize

jknipper:


pkg/controller/launch/pool_manager.go
```diff
@@ -126,7 +128,7 @@ func removeNodePool(pool []models.NodePoolInfo, name string) ([]models.NodePoolI
 }
 
 func (cpm *ConcretePoolManager) SetStatus(status *PoolStatus) error {
-	healthy, schedulable := cpm.healthyAndSchedulable()
+	healthy, schedulable, _ := cpm.healthyAndSchedulable([]string{})
```
Nuckal777, line 131: Why don't we pass `status.Nodes`?

Nuckal777 (commented):
lgtm

pkg/controller/launch/pool_manager.go
```diff
@@ -281,25 +283,54 @@ func (cpm ConcretePoolManager) unNeeded(nodes []openstack_kluster.Node) int {
...
+	schedulable := 0
+
 	for _, node := range nodes {
 		//Does the node belong to this pool?
 		if util.IsKubernikusNode(node.Name, cpm.Kluster.Spec.Name, cpm.Pool.Name) {
-			if !node.Spec.Unschedulable {
+			if node.Spec.Unschedulable {
+				if nodeIDs != nil && len(nodeIDs) > 0 {
+					id := strings.Replace(node.Spec.ProviderID, "openstack:///", "", 1)
+					pos := findString(nodeIDs, id)
```
uwe-mayer, line 306: I found this [IndexFunc()](https://pkg.go.dev/golang.org/x/exp/slices#IndexFunc) which should do the job?

pkg/controller/launch/pool_manager.go
jknipper: The ordered nodes are not used there. I wanted to avoid sorting stuff that is not needed later on. Maybe a bit ambiguous.

pkg/controller/launch/pool_manager.go
jknipper: Hmm, you need to pass a function for comparison, then it would be the same.

jknipper: Index would be the right one, no: https://cs.opensource.google/go/x/exp/+/c95f2b4c:slices/slices.go;l=106

databus23:
The sorting favour nodes that are not schedulable over nodes that are not even registered with kubernetes. I don't think if this is what we want. I would think that the order for removal should be:


* nodes not known to k8s
* nodes that are notschedulable and/or notready
* other nodes

uwe-mayer:
> The sorting favours nodes that are not schedulable over nodes that are not even registered with kubernetes. I don't think if this is what we want. I would think that the order for removal should be:
> 
> * nodes not known to k8s
> * nodes that are notschedulable and/or notready
> * other nodes

Does this sorting do it? @databus23 

```
sort.Slice(nodes, func(i, j int) bool {
		return 
		!util.IsKubernikusNode(nodes[i].Name, cpm.Kluster.Spec.Name, cpm.Pool.Name 
		|| nodes[i].Spec.Unschedulable) 
		|| i < j
	})
```

I think this needs 
```
nodeLister, err := cpm.nodeObservatory.GetListerForKluster(cpm.Kluster)
	if err != nil {
		return 0, 0, nil
	}

	nodes, err := nodeLister.List(labels.Everything())
	if err != nil {
		return 0, 0, nil
	}

```

jknipper:
Superseeded by #806

PR #582, opened 2021-07-13
Add support for Kubernetes 1.21

jknipper:
Adds support for Kubernetes 1.21 to Kubernikus. Also updates Cinder CSI components.


jknipper:
Dual stack support is now enabled by default in 1.21, see https://kubernetes.io/docs/concepts/services-networking/dual-stack/
We should take a close look if this causes any problems and needs to be disabled.

PR #585, opened 2021-07-19
Install latest rkt with ignition

databus23:
rkt is no longer shipped with flatcar and we need to bring our own.

### DIFF
--- a/pkg/templates/node_1.20.go
+++ b/pkg/templates/node_1.20.go
@@ -7,10 +7,14 @@ passwd:
   users:
     - name:          core
       password_hash: {{ .LoginPassword }}
+      groups: [ rkt ]
 {{- if .LoginPublicKey }}
       ssh_authorized_keys:
         - {{ .LoginPublicKey | quote }}
 {{- end }}
+  groups:
+    - name: rkt
+      system: true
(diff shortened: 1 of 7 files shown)


databus23:
Adding the `core` user to the `rkt` group does not work somehow.

pkg/templates/node_1.20.go
```diff
@@ -7,10 +7,14 @@ passwd:
   users:
     - name:          core
       password_hash: {{ .LoginPassword }}
+      groups: [ rkt ]
```
databus23, line 10: This does not work. its stripped out of the transpiled gobbeldigock that gets put into user-data:

```
  "passwd": {
    "users": [
      {
        "name": "core",
        "passwordHash": "$6$rounds=1000000$EHQwRTtoqLxuJNOM$/sU35OnW0cobfq1P8I6qQ/2DMqUNdpeYVzJjX8/NCPMkG.Q.F7lK0MfDRQWOgbE.jaFYzXsJVQwQGjbn/c3G11"
      }
    ],
    "groups": [
      {
        "name": "rkt",
        "system": true
      }
    ]
  }
```

PR #597, opened 2021-08-02
limit ssh key field to 10k, ensure userdata stays < 60k bytes

databus23:
This PR limits the maximum length of the ssh key field in the kubernikus api to 10k bytes which should allow to add at least 10 ssh keys to a cluster.
I also removed the json indenting for the ignition output and added a test that all ignition templates stay below 60k bytes (nova allows up to 65k) with a maxed out ssh key field.


BugRoger (approved):
LGTM

SchwarzM (approved):
lgtm

PR #600, opened 2021-08-03
Add support for custom root disks

databus23:
This PR adds a `customRootDiskSize` field to the nodepool spec. If set the vms get an ephemeral cinder based root disk that is deleted when the vm is deleted.

Has been requested a couple of times and was quite easy to implement. We can also use this for our concourse installation.

Example vm with a `customRootDiskSize: 1024`:
```
core@kks-class-pool1-6qv4n ~ $ df -h
Filesystem       Size  Used Avail Use% Mounted on
devtmpfs         959M     0  959M   0% /dev
tmpfs            985M     0  985M   0% /dev/shm
tmpfs            394M  5.5M  389M   2% /run
tmpfs            4.0M     0  4.0M   0% /sys/fs/cgroup
/dev/sda9        990G  1.3G  948G   1% /                    <--- :)
/dev/mapper/usr  985M  764M  171M  82% /usr
none             450M  312M  139M  70% /run/torcx/unpack
tmpfs            985M     0  985M   0% /media
tmpfs            985M     0  985M   0% /tmp
/dev/sr0         910K  910K     0 100% /media/configdrive
/dev/sda6        108M  108K   99M   1% /usr/share/oem
tmpfs            197M     0  197M   0% /run/user/500
core@kks-class-pool1-6qv4n ~ $
```

### DIFF
--- a/pkg/api/models/node_pool.go
+++ b/pkg/api/models/node_pool.go
@@ -25,6 +25,11 @@ type NodePool struct {
 	// config
 	Config *NodePoolConfig `json:"config,omitempty"`
 
+	// Create servers with custom (cinder based) root disked. Size in GB
+	// Maximum: 1024
+	// Minimum: 64
+	CustomRootDiskSize int64 `json:"customRootDiskSize"`
+
 	// flavor
 	// Required: true
 	Flavor string `json:"flavor"`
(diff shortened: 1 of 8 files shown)


jknipper (approved):
I tested various cases and the upgrade case. Also root disk volumes got deleted on server deletion. Looks good to me!

PR #601, opened 2021-08-06
flightcontroller: ensure more security group rules

databus23:
This PR refactors the `EnsureKubernikusRuleInSecurityGroup` to support reconciling more then one rule.
It now also creates the rule lazily only if no other broader rules are already covering the wanted rules.

In addition to the podCIDR rule we had, it now also added egress rules for ntp and the cluster apiserver.

Missing: regional swift and keppel for pulling images


pkg/client/openstack/kluster/client.go
```diff
@@ -190,94 +202,249 @@ func (c *klusterClient) SetSecurityGroup(sgName, nodeID string) (err error) {
...
 	}
 
-	groups, err := securitygroups.ExtractGroups(page)
+	return nil, err
+}
+
+func createLoadbalancerSecurityGroup(kluster *v1.Kluster, client *gophercloud.ServiceClient) (*securitygroups.SecGroup, error) {
+	sg, err := securitygroups.Create(client, securitygroups.CreateOpts{
+		Name:        fmt.Sprintf(LoadBalancerSecurityGroupName, kluster.Spec.Name),
+		Description: fmt.Sprintf(`Kubernikus: SecurityGroup containing LoadBalancer Self-IPs for cluster "%s"`, kluster.Spec.Name),
```
databus23, line 235:
```suggestion
		Description: fmt.Sprintf(`Kubernikus: SecurityGroup containing LoadBalancer VIP and Self-IPs for cluster "%s"`, kluster.Spec.Name),
```

pkg/client/openstack/kluster/client.go
```diff
@@ -190,94 +202,249 @@ func (c *klusterClient) SetSecurityGroup(sgName, nodeID string) (err error) {
...
+			for _, port := range allPorts {
+				if lb.VipPortID == port.ID {
+					found = true
+					break
+				}
+			}
+			if !found {
+				_, err := ports.Update(c.NetworkClient, lb.VipPortID, ports.UpdateOpts{SecurityGroups: &sg}).Extract()
+				if err != nil {
+					return false, err
```
databus23, line 292:
```suggestion
					return false, fmt.Errorf("Failed to add security group %s to port %s: %w", err, lbGroup.ID, lb.VipPortID, err)
```

pkg/client/openstack/kluster/client.go
```diff
@@ -190,94 +202,249 @@ func (c *klusterClient) SetSecurityGroup(sgName, nodeID string) (err error) {
...
+		return false, err
+	}
 
-		if rule.Protocol == string(rules.ProtocolTCP) {
-			tcp = true
-			continue
+	sg := []string{lbGroup.ID}
+	for _, lb := range allLoadbalancers {
+		if strings.HasPrefix(lb.Name, fmt.Sprintf("kube_service_%s", kluster.Spec.Name)) {
+			found := false
```
databus23, line 282: I suggest to replace the inner loop with

```
if !portsOfSecurityGroup.Has(lb.VipPortID) {
  //update port 
}
```
And to this before the whole loop construct:
```
portsOfSecurityGroup := sets.NewString()
for _, p := range allPorts {
  portsOfSecurityGroup.Insert(p.ID)
}
```
using https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets

pkg/client/openstack/kluster/client.go
```diff
@@ -190,94 +202,249 @@ func (c *klusterClient) SetSecurityGroup(sgName, nodeID string) (err error) {
...
+			EtherType:      string(rules.EtherType4),
+			Protocol:       string(rules.ProtocolTCP),
+			PortRangeMin:   443,
+			PortRangeMax:   443,
+			RemoteIPPrefix: apiIP.String(),
+			Description:    fmt.Sprintf(`Kubernikus: allow access to apiserver of cluster "%s"`, kluster.Spec.Name),
+		},
+	}
+
+	if lbGroup != nil && lbGroup.ID != "" {
```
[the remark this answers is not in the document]
jknipper: This can be nil, eg. when there is not enough quota in the project for creating an additional security group. This ensures the other rules even if that's the case.

pkg/client/openstack/kluster/client.go
```diff
@@ -190,94 +202,249 @@ func (c *klusterClient) SetSecurityGroup(sgName, nodeID string) (err error) {
...
 
-	udp := false
-	tcp := false
-	icmp := false
-	for _, rule := range groups[0].Rules {
-		if rule.Direction != string(rules.DirIngress) {
-			continue
+	if lbGroup == nil {
+		lbGroup, err = createLoadbalancerSecurityGroup(kluster, c.NetworkClient)
+		if err != nil || lbGroup == nil {
```
databus23, line 253: not sure the `|| lbGroup == nil` is required. Looking at createLoadbalancerSecurityGroup it returns an error when the create call fails. I don't see how `err` and `lbGroup` can be nil at the same time.

pkg/client/openstack/kluster/client.go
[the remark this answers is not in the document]
databus23: I see, good point but. it looks to me like at the moment at least if creating the security group fails there is an early return and no other security groups are enforced anyways.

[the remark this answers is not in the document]
jknipper: Where is it returning? I think I'm ignoring the error while getting the security group.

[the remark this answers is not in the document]
databus23: Here: https://github.com/sapcc/kubernikus/pull/601/files#diff-b5e03987e634a7e5e1033109b07cba3c54a48db540f4694445c66794899fdabfR255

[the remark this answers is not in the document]
jknipper: Isn't this only exiting `EnsureLoadbalancerSecurityGroup()`? To me it looks like `EnsureKubernikusRulesInSecurityGroup()` runs anyway.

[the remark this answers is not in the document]
databus23: ah yes, sorry I misread the code. Nvm

pkg/client/openstack/kluster/client.go
```diff
@@ -190,94 +202,249 @@ func (c *klusterClient) SetSecurityGroup(sgName, nodeID string) (err error) {
...
+}
+
+func (opts portListOpts) ToPortListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	params := q.Query()
+	q = &url.URL{RawQuery: params.Encode()}
+	return q.String(), err
+}
+
+func getLoadbalancerSecurityGroup(kluster *v1.Kluster, client *gophercloud.ServiceClient) (*securitygroups.SecGroup, error) {
```
databus23, line 217: As we are getting two different security groups in this controller maybe its worthwhile to have a single `getSecurityGroupByName(name string, client *gophercloud.ServiceClient)` function instead if this special purpose function?

jknipper: Yeah, makes sense.

databus23 (commented):
LGTM now. Would be awesome now change the e2e test to actually test that all this works:

* use an empty security group for e2e cluster, cleanup during teardown
* add loadbalancer happy path url check to test loadbalancer <-> nodes connectivity is working.

test/e2e/setup_test.go
```diff
@@ -37,6 +39,12 @@ func (s *SetupTests) Run(t *testing.T) {
 	require.True(t, running, "The Kluster must be Running")
 }
 
+func (s *SetupTests) CreateSecurityGroup(t *testing.T) {
+	sg, err := securitygroups.Create(s.OpenStack.Network, securitygroups.CreateOpts{Name: "kubernikus"}).Extract()
```
databus23, line 43: I would create a dedicated security group that is named after the current soak test cluster.

databus23 (commented):
Cool. One small nitpick but looks good otherwise

sandzwerg:
At least for NTP a customer had this issue today (not in a kubernikus environment) so I assume that the ingress rules are needed as well, but I haven't tested it myself. Here is the discussion: https://convergedcloud.slack.com/archives/C374AQJ3W/p1656505078940829

PR #602, opened 2021-08-09
Improve server listing

databus23:
Instead of listing all servers of a project we use a name filter the uses the regex `^(kks?)-$KLUSTERNAME-$POOLNAME` for the `name` attribute.
In addition we also process the pages individually when filtering through the returned servers.
This should avoid listing thousands of servers from s4 projects on a constant basis.

CAREFUL: This needs extra special testing and careful review. If the listing is somehow filtering out valid kubernikus nodes launchctl will spawn new servers until the quota is exceeded.


pkg/client/openstack/kluster/client.go
```diff
@@ -137,31 +137,26 @@ func (c *klusterClient) RebootNode(id string) error {
 }
 
 func (c *klusterClient) ListNodes(k *v1.Kluster, pool *models.NodePool) ([]Node, error) {
-	var unfilteredNodes []Node
 	var filteredNodes []Node
 
-	err := servers.List(c.ComputeClient, servers.ListOpts{}).EachPage(func(page pagination.Page) (bool, error) {
+	err := servers.List(c.ComputeClient, servers.ListOpts{Name: "^(kks-)?" + k.Spec.Name + "-" + pool.Name + "-"}).EachPage(func(page pagination.Page) (bool, error) {
```
databus23, line 142: @joker-at-work Is using a regex for listing servers by name safe and won't break in the future in your opionen? Seems like its depended on the database used for nova which unsettles me a bit.

joker-at-work: All I can say is that we don't plan to change the DB under Nova once again. 

FYI this is the function translating the string into the query: https://github.com/openstack/nova/blob/master/nova/db/sqlalchemy/api.py#L1943-L1973 Looks pretty DB-dependent. It basically just uses a sanitized version of whatever you provide.

databus23: Ok thanks.

PR #604, opened 2021-08-16
Fix Apiserver liveness check

databus23:
The apiserver liveness check was broken and not working.

Primarily `if _, err := os.Stat(LastFile); os.IsExist(err)` was incorrect for checking if a file exists.
But while debugging this I noticed that the overall logic was flawed as the metric only appears after the first restart.

This let me to rewrite the logic completely.

TODO: publish not apiserver images for 1.19/1.20 and update images.yaml


jknipper:
I added images with the new liveness probe and tested it, looks good! I also disabled support for buggy 1.19/1.20 releases.

PR #605, opened 2021-08-17
Don't scrape etcd sidecar twice

databus23:
Question for @auhlig how would we scrape the etcd metrics as well if we every decide do that? :)

### DIFF
--- a/charts/kube-master/charts/etcd/templates/deployment.yaml
+++ b/charts/kube-master/charts/etcd/templates/deployment.yaml
@@ -84,11 +84,6 @@ spec:
         app: {{ include "fullname" . }}
         release: {{ .Release.Name }}
         component: etcd
-{{- if .Values.backup.enabled }}
-      annotations:
-        prometheus.io/scrape: "true"
-        prometheus.io/port: "8080"
-{{- end }}
     spec:
       volumes:
         - name: data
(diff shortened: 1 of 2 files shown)


auhlig (approved):

auhlig:
Sorry. I missed this. 
This PR should fix the double scraping which I thought we already did a while ago. 
Currently checking whether I can fix the pod SD in k-qa-de-1

auhlig:
I don't know how we "fixed" this last time and this seems to be a ~shit show~ general problem with multi-container/port pods. 
The common practice is to either using the Service/[ServiceMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#servicemonitor) or a [PodMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmonitor) to explicitly specify the ports that should be scraped.

The relevant part is this:
prometheus.io annotations are configured on pod level and thus are valid for all containers.
The pod SD keeps containers of a pod if the port is 9102/metrics or if there's a prometheus.io/port annotation. [1]
However, I'm not able to configure that the container must expose a port matching `prometheus.io/port`. For that we'd require the Service or the monitors.
I think before we dropped the container name, so actually 2 targets were discovered but de-duplicated as the labels were the same.

[1] Relevant prometheus config:
```
- action: keep
  regex: (9102;.*;.*)|(.*;metrics;.*)|(\d+;.*;\d+)|(.*;.+;\d+)
  sourceLabels:
  - __meta_kubernetes_pod_container_port_number
  - __meta_kubernetes_pod_container_port_name
  - __meta_kubernetes_pod_annotation_prometheus_io_port
```

PR #611, opened 2021-09-07
Add static socat to apiserver image

databus23:
Haven't tested if this actually works with spore drill but the socat binary can be executed inside the container. Needs some manual testing and ultimately a bump in the images.yaml


jknipper (approved):
I tested this and it looks good. I also added new images with the socat binary.

PR #613, opened 2021-09-09
WIP: Enable basic k8s audit logging

databus23:
Logging is performed to stdout so it can be shipped with the normal log shipping we use in our clusters.

Open questions:
* refine the policy of what is logged based on tests in a bigger cluster (e.g. scaleout)
* enable by default would mean we generate log output for all customer clusters which might be heavily used and put a burden on our logging stack
  * Maybe we should log to file and upload to swift instead?


databus23:
Superseded by #629

PR #614, opened 2021-09-14
Switch to golangci-lint

databus23:
gometalinter is deprecated for a while now.

golangci-lint contains more linters and also included gofmt,goimmports and exportloopref so no need to run those seperatly 

~I'm not sure how to test the GitHub action thingy. I just copied the example from golanglint-ci~ Tested it and it works!

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -51,6 +51,7 @@ jobs:
       with:
         # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
         version: v1.42.1
+        args: -v --timeout 5m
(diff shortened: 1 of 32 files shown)


jknipper:
For whatever reason the affected filename is not displayed in gh actions, see https://github.com/sapcc/kubernikus/pull/617/checks?check_run_id=3673344063
Is it because of the colored output? Maybe turn it of by doing `--color=never`.

jknipper:
It seems there is a specific output option for gh actions:
```
output:
  # colored-line-number|line-number|json|tab|checkstyle|code-climate|junit-xml|github-actions
  # default is "colored-line-number"
  format: github-actions
```

PR #618, opened 2021-09-28
[WIP] Fix templates for cgroups v2 updates

jknipper:


pkg/templates/node_1.10.go
```diff
@@ -106,6 +113,9 @@ systemd:
       contents: |
         [Unit]
         Description=Kubelet via Hyperkube ACI
+        StartLimitAction=reboot
```
databus23, line 116: I'm a little concerned that this will cause reboots down the road then the kubelet is crashing for other reasons. Maybe we should create a dedicated unit that only crashes when the current kernel parameters are not set correctly or the cgroup support is loaded in the wrong version.

pkg/templates/node_1.19.go
```diff
@@ -43,6 +43,7 @@ systemd:
           contents: |
             [Service]
             Environment="DOCKER_OPTS=--log-opt max-size=5m --log-opt max-file=5 --ip-masq=false --iptables=false --bridge=none"
+            Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=cgroupfs"
```
databus23, line 46: do we need this for >= 1.19. Shouldn't the default work?

jknipper: Yes, it is needed. Default is now systemd for docker.

pkg/templates/node_1.10.go
jknipper: Good idea, but how to find out if it is loaded or not?

jknipper: Something like this should work:
```
    - name: legacy-cgroup-reboot.service
      enable: true
      contents: |
        [Unit]
        Description=Reboot if legacy cgroups are not enabled yet
        FailureAction=reboot

        [Service]
        Type=simple
        ExecStart=/usr/bin/grep 'cgroup' /proc/cmdline

        [Install]
        WantedBy=multi-user.target
```

pkg/templates/node_1.19.go
jknipper: Or configure the kubelet: https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/#migrating-to-the-systemd-driver

pkg/templates/node_1.10.go
jknipper: But I assume this will break current stable nodes.

PR #622, opened 2021-10-06
Fix resolv.conf race condition

### LINKED ISSUE #607: Race condition in node rkt based systems unit files breaks dns resolution
It seems like unit files that use rkt might start before dns is fully setup on the node leaving the started containers with an empty `/etc/resolv.conf`.

This was reported by gardener: https://github.tools.sap/kubernetes-staging/issues-staging/issues/16#issuecomment-320256

> Our current assumption is, that the root cause is a race condition on node boot/restart between the systemd services kube-proxy and systemd-resolved.
> When starting kube-proxy rkt uses the host's /etc/resolv.conf for the container as well (see flag --dns=host).
However, this file on the host is managed by systemd-resolved and it might not be filled yet by the time, that the kube-proxy service starts.
If this assumption holds true one fix could be to use the systemd After directive to express that the kube-proxy systemd service starts after the systemd-resolved service is running.

It seems like when `rkt` is used with `--dns=host` the `/etc/resolv.conf` is copied during startup of the container and never updated. :( So we need to make sure that the unit either starts after dns setup is complete or fails and relies and systemd to restart if the resolv.conf is empty.

jknipper:
This hopefully fixes #607

### DIFF
--- a/pkg/templates/ignition.go
+++ b/pkg/templates/ignition.go
@@ -27,7 +27,7 @@ var Ignition = &ignition{}
 
 var passwordHashRounds = 1000000
 
-const TEMPLATE_VERSION = "2"
+const TEMPLATE_VERSION = "1"
 
 func (i *ignition) getIgnitionTemplate(kluster *kubernikusv1.Kluster) (string, error) {
 	switch {
--- a/pkg/templates/ignition.go
+++ b/pkg/templates/ignition.go
@@ -27,7 +27,7 @@ var Ignition = &ignition{}
 
 var passwordHashRounds = 1000000
 
-const TEMPLATE_VERSION = "1"
+const TEMPLATE_VERSION = "3"
 
 func (i *ignition) getIgnitionTemplate(kluster *kubernikusv1.Kluster) (string, error) {
 	switch {
--- a/pkg/templates/node_1.10.go
+++ b/pkg/templates/node_1.10.go
@@ -65,6 +65,7 @@ systemd:
         Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/flatcar/flannel-wrapper.uuid"
         EnvironmentFile=-/run/flannel/options.env
 
+        ExecStartPre=/usr/bin/host identity-3.{{ .OpenstackRegion }}.cloud.sap
         ExecStartPre=/sbin/modprobe ip_tables
         ExecStartPre=/usr/bin/mkdir --parents /var/lib/flatcar /run/flannel
         ExecStartPre=-/opt/bin/rkt rm --uuid-file=/var/lib/flatcar/flannel-wrapper.uuid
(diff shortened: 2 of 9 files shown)


Nuckal777:
Long shot for the actual issue: [This](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/) mentions in the "Cut the crap! How do I make sure that my service starts after the network is really online?" section, that for the network-online.target a proper wait service needs to be enabled. Maybe that one (systemd-networkd-wait-online.service) is disabled?

databus23:
@Nuckal777 Interesting. I spot check one kubernikus node and:

```
core@node01~ $ systemctl is-enabled  systemd-networkd-wait-online.service
disabled
```

Ilso are we sure the `network-online.target` also means "dns is configured"?

databus23:
so
```
> systemctl status systemd-networkd-wait-online.service
● systemd-networkd-wait-online.service - Wait for Network to be Configured
     Loaded: loaded (/usr/lib/systemd/system/systemd-networkd-wait-online.service; disabled; vendor preset: disabled)
     Active: active (exited) since Mon 2021-09-27 15:04:07 UTC; 1 week 2 days ago
       Docs: man:systemd-networkd-wait-online.service(8)
   Main PID: 1128 (code=exited, status=0/SUCCESS)

Sep 27 15:04:07 ... systemd[1]: Starting Wait for Network to be Configured...
Sep 27 15:04:07 ... systemd[1]: Finished Wait for Network to be Configured.
```

Nuckal777:
I expect that is not the issue. 😢 The [ArchWiki](https://wiki.archlinux.org/title/Systemd#Running_services_after_the_network_is_up) mentions to include `After=network-online.target nss-lookup.target` for DNS to be available. Have we tried that?

SuperSandro2000:
I did a "quick" (it took over 2 minutes to run those commands) ``systemd-analyze dot | dot -Tsvg -o dot.svg`` and ``systemd-analyze plot > plot.svg`` to figure out which things start after network-online.target on one machine I have access to. One example is wireguard which does the following in its service:

```
cat /lib/systemd/system/wg-quick@.service
[Unit]
Description=WireGuard via wg-quick(8) for %I
After=network-online.target nss-lookup.target
Wants=network-online.target nss-lookup.target
PartOf=wg-quick.target
Documentation=man:wg-quick(8)
Documentation=man:wg(8)
Documentation=https://www.wireguard.com/
Documentation=https://www.wireguard.com/quickstart/
Documentation=https://git.zx2c4.com/wireguard-tools/about/src/man/wg-quick.8
Documentation=https://git.zx2c4.com/wireguard-tools/about/src/man/wg.8

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/bin/wg-quick up %i
ExecStop=/usr/bin/wg-quick down %i
ExecReload=/bin/bash -c 'exec /usr/bin/wg syncconf %i <(exec /usr/bin/wg-quick strip %i)'
Environment=WG_ENDPOINT_RESOLUTION_RETRIES=infinity

[Install]
WantedBy=multi-user.target
```

So I think we should set the following which is very similar to the suggestion from Erik from the archwiki to get what we want:

```
[Unit]
After=network-online.target nss-lookup.target
Wants=network-online.target nss-lookup.target
```

PR #629, opened 2021-11-25
Audit logging for API server

Nuckal777:
This PR enables the API server's audit logs and ships them using fluentd to elasticsearch or swift, conveniently controlled by the `audit` API flag. Elasticsearch credentials have to provided as a secret in the API server's namespace. The swift backend uses the Kubernikus kluster service user.

Open questions:
- It might be useful to set a limit for the `emptyDir` sizes.
- The swift backend currently uploads audit logs every 5min, some other interval could be appropiate.


charts/kube-master/templates/api.yaml
```diff
@@ -121,6 +121,16 @@ spec:
             items:
               - key: local-kubeconfig
                 path: kubeconfig
+        {{- if ne .Values.audit "" }}
```
databus23, line 124: Can be simplified to
```suggestion
        {{- if Values.audit }}
```

charts/kube-master/templates/api.yaml
```diff
@@ -244,6 +254,12 @@ spec:
             - --oidc-username-prefix=-
             - --oidc-username-claim=name
             {{ end }}
+            {{- if ne .Values.audit "" }}
```
databus23, line 257:
```suggestion
            {{- if Values.audit }}
```

charts/kube-master/templates/api.yaml
```diff
@@ -265,6 +281,13 @@ spec:
             - mountPath: /liveness-probe
               name: liveness-probe
 {{- end }}
+            {{- if ne .Values.audit "" }}
```
databus23, line 284:
```suggestion
            {{- if .Values.audit }}
```

charts/kube-master/templates/api.yaml
```diff
@@ -320,3 +343,69 @@ spec:
             - mountPath: /etc/kubernetes/config
               name: wormhole-config
               readOnly: true
+        {{- if ne .Values.audit "" }}
```
databus23, line 346:
```suggestion
        {{- if .Values.audit }}
```

charts/kube-master/templates/configmap.yaml
```diff
@@ -85,3 +85,99 @@ data:
     except requests.exceptions.RequestException as e:
       print e
       sys.exit(0)
+{{- if ne .Values.audit "" }}
```
databus23, line 88:
```suggestion
{{- if .Values.audit }}
```

charts/kube-master/templates/configmap.yaml
```diff
@@ -85,3 +85,99 @@ data:
...
+  audit-policy.yaml: |-
+    apiVersion: audit.k8s.io/v1
+    kind: Policy
+    rules:
+    - level: None
+      userGroups:
+      - system:nodes
+    - level: None
+      users:
+      - gardener
```
databus23, line 98: I think we can drop all the users with garden in the name. This only applies to gardener clusters

pkg/client/openstack/admin/client.go
```diff
@@ -21,13 +21,14 @@ import (
 	"github.com/sapcc/kubernikus/pkg/client/openstack/roles"
 )
 
-var serviceUserRoles = []string{"network_admin", "member"}
+var serviceUserRoles = []string{"network_admin", "member", "swiftoperator"}
```
databus23, line 24: We should not assign the `swiftoperator` role to the kubernikus service user. This allows read/write access to all swift data in the project. Instead we need to create the audit container and give the user access to this container only, using the swift acls.
Are are already doing this for the etcd backup using the admin client here: https://github.com/sapcc/kubernikus/blob/3d768210318ba91d10b3f1746f29520c39f62492/pkg/client/openstack/admin/client.go#L346

I'm not sure if we are reconciling the swift container but if we don't I would suggest that we do that for both containers in ground control. (E.g. in the running phase if the etcd backup is enabled we make sure the storage container is (re)created, if the audit logging is enabled we (re)create the audit container.

databus23 (changes requested):
Thanks for your work! I left a few remarks

pkg/apis/kubernikus/v1/secret.go
```diff
@@ -16,6 +16,9 @@ type Secret struct {
 	DexClientSecret   string `json:"dex-client-secret,omitempty"`
 	DexStaticPassword string `json:"dex-static-password,omitempty"`
 
+	AuditUserDomainID    string `json:"audit-user-domain-id,omitempty"`
```
databus23, line 19: I'm wondering why this is prefixed with `Audit`. Isn't this the domain_id of the kubernikus service user and the domain_id of the hosting project? So this would be rather generic and should be called `UserDomainID` and `ProjectDomainID`?

databus23, line 19: I'm wondering why this is prefixed with Audit. Isn't this the domain_id of the kubernikus service user and the domain_id of the hosting project? So this would be rather generic and should be called UserDomainID and ProjectDomainID?

pkg/client/openstack/admin/client.go
```diff
@@ -376,6 +378,31 @@ func (c *adminClient) CreateStorageContainer(projectID, containerName, serviceUs
...
+
+	storageClient, err := openstack.NewObjectStorageV1(c.ProviderClient, gophercloud.EndpointOpts{})
+	if err != nil {
+		return false, err
+	}
+	storageClient.Endpoint = endpointURL
+
+	_, err = containers.Get(storageClient, containerName, containers.GetOpts{}).Extract()
+	if err != nil {
+		err404 := gophercloud.ErrDefault404{}
```
databus23, line 395: I think this can be simplified using a type assertion as we do here:
https://github.com/sapcc/kubernikus/blob/master/test/e2e/pyrolisis_test.go#L133

charts/kube-master/templates/configmap.yaml
```diff
@@ -85,3 +85,93 @@ data:
...
+      users:
+      - kubelet
+      - system:kube-controller-manager
+      - system:kube-scheduler
+      - system:kube-aggregator
+      - system:kube-proxy
+      - system:apiserver
+      - system:serviceaccount:kube-system:generic-garbage-collector
+      - system:serviceaccount:kube-system:namespace-controller
+      - garden.sapcloud.io:monitoring
```
databus23, line 106: those garden.sapcloud.io references can go as well

pkg/api/models/kluster_spec_audit.go
```diff
@@ -0,0 +1,52 @@
+// Code generated by go-swagger; DO NOT EDIT.
```
databus23, line 1: I think this whole file is not required anymore. It seems like its from your attempt to include the audit config in the kluster spec, no?

pkg/client/openstack/admin/client.go
```diff
@@ -376,6 +378,31 @@ func (c *adminClient) CreateStorageContainer(projectID, containerName, serviceUs
 	return err
 }
 
+func (c *adminClient) ExistsStorageContainer(projectID, containerName string) (bool, error) {
```
databus23, line 381: Maybe `StorageContainerExists` ? The other way around does not roll off the tongue that easily.

swagger.yml
```diff
@@ -467,6 +467,9 @@ definitions:
     properties:
       openstack:
         $ref: '#/definitions/OpenstackSpec'
+      audit:
```
databus23, line 470: if we make this an enum, the docs will tell what values are valid as we do for backup: https://github.com/sapcc/kubernikus/blob/master/swagger.yml#L526
wdyt?

databus23 (changes requested):
This is looking good. I have a few more nitpicks. Thanks!

databus23 (approved):
LGTM. Thanks!

PR #631, opened 2021-12-20
Audit logging improvements

Nuckal777:


charts/kube-master/templates/api.yaml
```diff
@@ -346,7 +346,11 @@ spec:
         {{- if .Values.audit }}
         - name: fluentd
           image: {{ include "fluentd.image" . | quote }}
-          imagePullPolicy: Always
+          resources:
+            requests:
+              ephemeral-storage: "2Gi"
+            limits:
+              ephemeral-storage: "2Gi"
```
databus23, line 353: Thinking about this some more I'm wondering if we really should set a hard limit here at this point. We don't really know the log volume in a productive landscape and how fast the space will grow once the receiving side is down for a few hours. Maybe its safer to just start with requests only and leave the limits for later when we have more experience with this.

databus23 (changes requested):
On second thought I have some concerns regarding the ephemeral storage limits. otherwise looks good.

PR #644, opened 2022-03-03
Skip node replacement if pool is already degraded

jknipper:


databus23:
Hmm this PR looks like it only modifies the `Reboot` case not the `Replace/Upgrade` case

jknipper:
> Hmm this PR looks like it only modifies the `Reboot` case not the `Replace/Upgrade` case

Ah, nice catch. I added it there as well.

databus23 (approved):

databus23:
@jknipper Can we merge this?

jknipper:
I need to take a look again if this aged well.

PR #645, opened 2022-03-14
Remove deprecated spec.template.spec.nodeSelector

vzabawski:
spec.template.spec.nodeSelector[beta.kubernetes.io/os] is deprecated since v1.14, so it might be good to replace it with `kubernetes.io/os`


databus23:
Thanks for bringing this up! It should be fixed now.

PR #648, opened 2022-03-30
Add support for Kubernetes 1.23

jknipper:
Adds support for Kubernetes 1.23. The following components has been upgraded:
- etcd 3.4.13
- etcd-backup-restore 0.15.1
- Cinder CSI components
- Openstack cloud controller manager 1.23.1
- CoreDNS 1.9.1
- Flannel 0.17.0
- CNI plugins 1.0.1
- fluentd v1.14.6


pkg/controller/ground/bootstrap/csi/csi123.go
```diff
@@ -0,0 +1,217 @@
...
+
+	err = createClusterRoleBinding(client, CSIClusterRoleBindingResizer)
+	if err != nil {
+		return errors.Wrap(err, "CSIClusterRoleBindingResizer")
+	}
+
+	err = createClusterRoleBinding(client, CSIClusterRoleBindingSnapshotter)
+	if err != nil {
+		return errors.Wrap(err, "CSIClusterRoleBindingSnapshotter")
+	}
```
Nuckal777, line 109: Can we just call `SeedCinderCSIRoles123` instead?

Nuckal777 (commented):
We may want to bump kube-master version. Everything else looks fine to me.

pkg/controller/ground/bootstrap/csi/csi123.go
jknipper: Yeah, but this is only a subset of changes used on kluster upgrade. I found this more confusing.

PR #649, opened 2022-04-21
Switch to keppel.global for container images

Nuckal777:
- [x] Switch to keppel.global for all regions to get a fallback in case keppel is down in the region.
- [ ] Switch all references to upstream images we currently vendor unchanged keppel/ccloud to use a proper keppel mirror (e.g. `keppel.eu-de-1.cloud.sap/ccloud-registry-k8s-io-mirror`)
  - kube-scheduler
  - kube-controller-manager
  - kube-proxy
  - openstack-cloud-controller-manger
  - csi-*
  - cinder-csi-plugin

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -2,826 +2,826 @@ imagesForVersion:
   '1.22.4':
     supported: true
     apiserver:
-      repository: 'keppel.$REGION.cloud.sap/ccloud/kube-apiserver'
+      repository: 'keppel.global.cloud.sap/ccloud/kube-apiserver'
       tag: 'v1.22.4'
     controllerManager:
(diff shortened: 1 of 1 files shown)


databus23 (approved):

databus23:
One last thought. Is this affecting the keppel image garbage collection. Like does gatekeeper still prevent keppel do gc images that are in active use in the cluster? @majewsky

majewsky:
Good thought, but it does not. GC policies are evaluated on the primary account in eu-de-1. Since we aggregate last-pulled events from the replica accounts into the primary, it does not matter where the image is pulled exactly.

databus23 (changes requested):
In adding to switching to keppel.global we should also switch to prober keppel mirrors for registry.k8s.io

Nuckal777:
Closing in favor of #719 .

PR #654, opened 2022-05-12
Add cgroupsV1 flag to nodepools

databus23:
This allows falling back to cgroups v1 for clusters > 1.18.

The node pool now has a `cgroupsV1` boolean flag to bring up nodes with cgroups v1 enabled

Fallingback to cgroupv1 is possible without a reboot since flatcar 3033.2.4: https://www.flatcar.org/docs/latest/container-runtimes/switching-to-unified-cgroups/

### DIFF
--- a/pkg/templates/node_1.19.go
+++ b/pkg/templates/node_1.19.go
@@ -43,6 +43,9 @@ systemd:
           contents: |
             [Service]
             Environment="DOCKER_OPTS=--log-opt max-size=5m --log-opt max-file=5 --ip-masq=false --iptables=false --bridge=none"
+            {{- if .CGroupsV1 }}
+            Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=cgroupfs"
+            {{- end }}
     - name: flanneld.service
       enable: true
       contents: |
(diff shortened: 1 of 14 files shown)


pkg/templates/node_1.19.go
```diff
@@ -43,6 +43,9 @@ systemd:
           contents: |
             [Service]
             Environment="DOCKER_OPTS=--log-opt max-size=5m --log-opt max-file=5 --ip-masq=false --iptables=false --bridge=none"
+            {{- if .CGroupsV1 }}
+            Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=cgroupfs"
```
defo89, line 47: this line should be `Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=systemd"`
cc @SchwarzM

defo89, line 47: this line should be `Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=systemd"`
cc @SchwarzM

databus23:
Apparently this is not needed anymore as we don't have pressing bug reports about cgroupsv2 with recent kubernetes versions

PR #658, opened 2022-05-23
kube-master: support for networking.k8s.io/v1 ingress objects

databus23:


databus23:
Good catch. I bumped the chart.

Nuckal777 (approved):
🎉

PR #659, opened 2022-05-27
Fix ingress api version selection

databus23:
Turns out using  `.Capabilities.APIVersions.Has "networking.k8s.io/v1”` is flawed because while the api version is already available in 1.15 but does not contain the `Ingress` resource.

Using `.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress”` is only supported with helm >= helm 2.15 and we are still using helm 2.11 in k- controlplanes.

So we have to go by kubernetes version instead to decide if to use networking.k8s.io/v1 for ingress resources.

### DIFF
--- a/charts/kube-master/Chart.yaml
+++ b/charts/kube-master/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: kube-master
-version: 2.1.2
+version: 2.1.3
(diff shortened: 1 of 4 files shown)


Nuckal777:
Please bump the chart version. :blush:

databus23:
Done!

Nuckal777 (approved):

PR #662, opened 2022-06-13
Load rkt from regional objectstore

databus23:
This PR removes the dependency ob objectstore-3.eu-de-1.cloud.sap from all regions by fetching the rkt components from the regional objectstore instead.
This files a are the same ones used by the controlplane terraform scripts which are mirrored to all regions.


BugRoger (approved):
LGTM

PR #669, opened 2022-07-04
Use patch instead of update when bumping migration version

databus23:
This should avoid failed updates becuse of change resourced version on the kluster object


Nuckal777 (dismissed):
lgtm

databus23:
Superseded by #698

PR #672, opened 2022-07-07
Bump to k8s 1.18-> 1.21, helm 3.3-> 3.6

databus23:
* this allows compiling with go 1.18
* I used helm 3.6 because it depends on k8s 1.21 as well
* I removed the vendor folder, because why not?
* remove helm2


Nuckal777:
Kick out Helm2 as well?

databus23:
I serve at your leisure...

Nuckal777:
All deployments should be migrated. Even if not, we only need to pick a version that does the conversion and can then upgrade further.

Nuckal777 (approved):

PR #673, opened 2022-07-12
include scheduler config into extra vars

SchwarzM:
i want to be able to change the scheduler config for BM klusters


SchwarzM:
just coded

SchwarzM:
i've tested manually but no value interpolations done

databus23 (approved):

PR #675, opened 2022-07-22
Add release logging to pipeline

SuperSandro2000:


ci/pipeline.yaml.erb
```diff
@@ -433,13 +450,17 @@ jobs:
...
+          ensure:
+            put: <%= region %>.release
         - task: kubernikus-dex_<%= region %>
           config:
             <<: *task_helm_kubernikus-dex
           params:
             REGION: <%= region %>
             CONTEXT: k-<%= region %>
             <<: *auth_<%= region %>
+          ensure:
```
databus23, line 462: As said here: https://convergedcloud.slack.com/archives/C6HHGQZG8/p1659365292755689?thread_ts=1659365109.803789&cid=C6HHGQZG8 I would not do release logging for the kubernikus-dex deployment. It currently breaks the pipeline and instead of fixing it I would just remove it

databus23: I updated the PR to not try release logging for kubernikus-dex which unwedges the pipeline

databus23:
As this is already applied to the pipeline I'm going to merge this now.

PR #678, opened 2022-08-10
Add support for Kubernetes 1.22.12

Nuckal777:
This includes a fix for https://github.com/kubernetes/kubernetes/issues/105204 which a customer stumbled upon. To my knowledge the issue does not break anything but is very misleading. I believe the version of 1.23 we deliver is affected as well.

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -73,8 +73,82 @@ imagesForVersion:
     fluentd:
       repository: 'keppel.$REGION.cloud.sap/ccloud/kubernikus-fluentd'
       tag: 'v1.14.6-1.1'
-  '1.22.4':
+  '1.22.12':
     supported: true
+    apiserver:
+      repository: 'keppel.$REGION.cloud.sap/ccloud/kube-apiserver'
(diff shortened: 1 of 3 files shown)


charts/images.yaml
```diff
@@ -73,6 +73,80 @@ imagesForVersion:
...
+      repository: 'keppel.$REGION.cloud.sap/ccloud/cinder-csi-plugin'
+      tag: 'v1.22.0'
+    flannel:
+      repository: 'keppel.$REGION.cloud.sap/ccloud-quay-mirror/coreos/flannel'
+      tag: 'v0.12.0'
+    fluentd:
+      repository: 'keppel.$REGION.cloud.sap/ccloud/kubernikus-fluentd'
+      tag: 'v1.14-1'
   '1.22.4':
     supported: true
```
databus23, line 151: We should set this to `false` to have the version not show um in elektra anymore. Otherwise lgtm

databus23:
we wait for 1.22.14 from the pipeline

PR #680, opened 2022-08-22
Remove rkt (for 1.24)

databus23:
This PR removes rkt from the 1.24 ignition files.

* kubelet runs directly via systems without containerisation
* kube-proxy and wormhole run as daemonsets

requires #677 

Note to reviewer: As this is based on #677 its advised to focus the last commit: https://github.com/sapcc/kubernikus/pull/680/commits/17bd14e861ee77c0adf23b452d8d5aca0f44bca3


contrib/cni-plugins/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine
+ARG TARGETARCH
+ARG CNI_PLUGINS_RELEASE
+
+LABEL source_repository=https://github.com/sapcc/kubernikus/tree/master/contrib/cni-plugins
+WORKDIR /cni-plugins
+RUN mkdir -p /cni-plugins && \
+    wget -O- https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_RELEASE}/cni-plugins-linux-${TARGETARCH}-${CNI_PLUGINS_RELEASE}.tgz | tar -xz
```
defo89, line 8: @databus23 would it be ok to also add multus cni here?

defo89: nvm, created a separate image for it https://github.com/sapcc/kubernikus/commit/b5827c368b796c181b41d738c4bdcef99529b656

defo89:
finding from the metal world: need to add dropin, otherwise default config is used https://flatcar-linux.org/docs/latest/container-runtimes/customizing-docker/#use-a-custom-containerd-configuration

should also consider adding for seamless transition from docker
```
[plugins."io.containerd.grpc.v1.cri"]
enable_unprivileged_ports = true  <<
enable_unprivileged_icmp = true <<
```

pkg/controller/ground/bootstrap/network/kube_proxy_manifests.go
```diff
@@ -0,0 +1,180 @@
...
+      - name: proxy
+        image: "{{ .KubeProxy }}"
+        imagePullPolicy: IfNotPresent
+        lifecycle:
+          postStart:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - sleep 5
```
Nuckal777, line 134: I guess this can be removed?

databus23: Good question? Why is there in the first place. I copied the daemonset from https://github.com/sapcc/helm-charts/blob/master/system/kube-proxy-ng/templates/daemonset.yaml @defo89 Any insights? At a glance this lifecycle hook seems non sensical to me.

databus23:
> should also consider adding for seamless transition from docker

Thanks @defo89 . I updated the #677 PR which introduces containerd with custom config in the first place

pkg/controller/ground/bootstrap/network/kube_proxy_manifests.go
defo89: I don't remember the details anymore, apparently we had some finding in older kube-proxy version that it needs some delay to get the rules up to date https://github.com/sapcc/helm-charts/commit/ce01be9333bebce7102eca26ebf3409090112673

PR #700, opened 2022-10-06
Add k8s 1.25 support

jknipper:


defo89:
Hi @jknipper can we also include https://github.com/sapcc/kubernikus/pull/701? Thanks!

jknipper:
> Hi @jknipper can we also include #701? Thanks!

@defo89 I merged changes from master and 1.25 is using node template from 1.24 so it should be in.

PR #708, opened 2022-10-31
[wormhole] check for up to date cert

defo89:
* same as it is done in [wormhole client](https://github.com/sapcc/helm-charts/blob/master/system/wormhole/templates/daemonset.yaml)
* livenessProbe will fail on cert rotation
* new cert will be used after restart


charts/kube-master/templates/api.yaml
```diff
@@ -356,6 +358,15 @@ spec:
...
               readOnly: true
+          livenessProbe:
+            exec:
+              command:
+              - sh
+              - -c
+              - |
+                cmp -s /etc/kubernetes/certs/tls-wormhole.pem /tmp/tls-wormhole.livenessprobe
+            initialDelaySeconds: 60
+            periodSeconds: 60
```
databus23, line 369: we should set `timeoutSeconds: 3` to avoid the ominous timeouts we are seeing with the default setting of `1` in the last few months

databus23: And do we even need a shell invocation here or can we execute `cmp` directly? I think the original needed variable expansion which we don't do here

defo89: thanks @databus23 - I have added the increased `timeoutSeconds`. 
Tested executing `cmp` directly in qa labs - works fine, too.

charts/kube-master/templates/api.yaml
```diff
@@ -356,6 +358,13 @@ spec:
             - mountPath: /etc/kubernetes/config
               name: wormhole-config
               readOnly: true
+          livenessProbe:
+            exec:
+              command:
+              - cmp -s /etc/kubernetes/certs/tls-wormhole.pem /tmp/tls-wormhole.livenessprobe
```
databus23, line 364: you are sure this works?
I think this should be:
```
- cmp
- -s
- /etc/kubernetes/certs/tls-wormhole.pem
- /tmp/tls-wormhole.livenessprobe
```

or `[cmp, -s, /etc/kubernetes/certs/tls-wormhole.pem, /tmp/tls-wormhole.livenessprobe]`

defo89: sigh, of course you are right (facepalm)

charts/kube-master/templates/api.yaml
```diff
@@ -335,16 +335,18 @@ spec:
...
-            - --kubeconfig=/etc/kubernetes/config/kubeconfig
-            - --ca=/etc/kubernetes/certs/apiserver-nodes-ca.pem
-            - --cert=/etc/kubernetes/certs/tls-wormhole.pem
-            - --key=/etc/kubernetes/certs/tls-wormhole-key.pem
-            - --service-cidr={{ .Values.serviceCIDR }}
-            - --api-port={{ .Values.advertisePort }}
-            - --v=5
+          command:
+            - sh
+            - -c
```
databus23, line 340: final nitpick. We should exit on any errors that happen while copying the cert (e.g. can't be found)
```suggestion
            - -ec
```

databus23 (approved):

Issue #714, opened 2022-11-18
etcbackup sidecar warns about missing env var.


databus23:
`time="2022-11-18T15:03:32Z" level=warning msg="POD_NAMESPACE environment variable is not set. The variable is used to safe guard against wrong volume mount"`

Sounds like adding this is a low hanging fruit.

PR #715, opened 2022-11-18
Add pod namespace env to etcd backup sidecar

jknipper:
Addresses #714

### DIFF
--- a/charts/kube-master/charts/etcd/templates/deployment.yaml
+++ b/charts/kube-master/charts/etcd/templates/deployment.yaml
@@ -264,6 +264,10 @@ spec:
                   name: {{ include "fullname" . }}
                   key: openstack-project-id
 {{- end }}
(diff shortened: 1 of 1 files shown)


databus23:
What is this actually doing? I'm a bit confused why it needs the namespace now that I think about it. The namespace for our deployments will always by `kubernikus`. How does this prevent from mounting the wrong volume...

jknipper:
See https://github.com/gardener/etcd-backup-restore/blob/b68a54ea05c6d33a91adfcfd872b667357268303/pkg/initializer/validator/datavalidator.go#L93

Looks like they write a file with the namespace in it to the volume and verify it on startup.

databus23:
Closing after discussing in out dev call. We could add kluster name as `POD_NAMESPACE` but we are not sure why

PR #716, opened 2022-11-18
wormhole: fix concurrent map access

databus23:
seen a few crashes in eu-de-1 with 200+ nodes


jknipper (approved):
Looks good to me!

Nuckal777:
Could we `defer` the unlocks? Adding additional branches may setup for deadlocks.

databus23:
@Nuckal777 I added defer statements for unlocking were possible

PR #717, opened 2022-11-18
launch controller cleanup needs to wait on deorbiter

databus23:
At least the volume cleanup in the deorbiter needs the csi daemonset on the nodes to unmount the volumes.

So we should just wait with the termination of nodes until the deorbiter is done


databus23:
Superseded by #720

PR #737, opened 2022-12-22
Update in-cluster images during cluster upgrade

### LINKED ISSUE #736: upgrading from 1.24.x leaves kube-proxy/flannel/wormhole ds at old version
Since we started to deploy kube-system, flannel and wormhole as daemonsets we forgot to handle the upgrade case so when upgrading from 1.24.X -> 1.24.y or 1.25.z the daemonsets stay as is and are not upgraded.

I guess we finally need to merge https://github.com/sapcc/kubernikus/pull/679

If this is not feasible we need to think about a stopgap solution and handle at least updating images during cluster upgrade for kube-proxy.

databus23:
This is a stopgap for #736  solution until we merge https://github.com/sapcc/kubernikus/pull/679

During cluster upgrade the image versions of the following in cluster daemonsets we bootstrap are updated according to `images.yaml`:

* kube-proxy
* wormhole
* cinder-csi-nodeplugin

Fixes #736

### DIFF
--- a/pkg/controller/ground.go
+++ b/pkg/controller/ground.go
@@ -36,8 +36,10 @@ import (
 	"github.com/sapcc/kubernikus/pkg/util"
 	etcd_util "github.com/sapcc/kubernikus/pkg/util/etcd"
 	helm_util "github.com/sapcc/kubernikus/pkg/util/helm"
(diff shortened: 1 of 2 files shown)


Nuckal777 (approved):
lgtm

PR #746, opened 2023-01-17
Add snapshot pyrolisis

uwe-mayer:
- Cleans up Snapshots in Pyrolisis:
  Had to do a little dance with the openstack go client, as it deletes volumes corresponding to snapshots.
- Moves Volume+Snapshot Cleanup before `WaitForE2EKlustersTerminated` as `TerminateCluster` doesn't clean up snapshots -> Volume deletion gets stuck -> Kluster termination timeouts (undo with Deorbiter update?)


test/e2e/pyrolisis_test.go
```diff
@@ -164,6 +168,48 @@ func (p *PyrolisisTests) CleanupBackupStorageContainers(t *testing.T) {
...
+	require.NoError(t, err, "There should be no error extracting all snapshots")
+
+	var volumesInDeletionIds = make([]string, 1)
+	for _, snapshot := range allSnapshots {
+		err := snapshots.Delete(storageClient, snapshot.ID).ExtractErr()
+		require.NoError(t, err, "There should be no error deleting the snapshot")
+		volumesInDeletionIds = append(volumesInDeletionIds, snapshot.VolumeID)
+	}
+
+	// The storage client automatically deletes volumes alongside snapshots.
```
databus23, line 195: What? If you delete a snapshot the volume is deleted as well? That does not make sense to me. What do you men by that.

uwe-mayer:
Added checks for:

- volume attached to snapshot was created in `e2e` test before snapshot pyrolisis
- volume is in stable state for deletion before volume pyrolisis

test/e2e/pyrolisis_test.go
```diff
@@ -164,6 +172,49 @@ func (p *PyrolisisTests) CleanupBackupStorageContainers(t *testing.T) {
...
+
+	allPages, err := snapshots.List(storageClient, listOpts).AllPages()
+	require.NoError(t, err, "There should be no error retieving all snapshot pages")
+
+	allSnapshots, err := snapshots.ExtractSnapshots(allPages)
+	require.NoError(t, err, "There should be no error extracting all snapshots")
+
+	for _, snapshot := range allSnapshots {
+
+		volume, volumeErr := volumes.Get(storageClient, snapshot.VolumeID).Extract()
```
databus23, line 194: instead of looking up metadata on the volume, use the metadata on the snapshot: e.g. `cinder.csi.openstack.org/cluster: e2e-gzd78-d93ad03198be4783befe14afad24c878`

The check should be something like `strings.HasPrefix("e2e-")`

uwe-mayer: Changed. Ready for merge?

PR #757, opened 2023-02-10
Add snapshot deletion to Deorbiter

uwe-mayer:
`e2e` tests run succesfully  --> Deorbiter successfully cleans up snapshots, therefore volumes can be deleted and CleanUp is successful


pkg/controller/deorbit/deorbiter.go
```diff
@@ -138,6 +170,20 @@ func (d *ConcreteDeorbiter) DeleteServices() (deleted []core_v1.Service, err err
...
+
+	if err != nil {
+		return err
+	}
+
+	if done {
+		return nil
+	}
+
+	return wait.PollUntil(PollInterval, d.isSnapshotCleanupFinished, d.Stop)
```
Nuckal777, line 184: Can be replaced with `wait.PollImmediateUntil()`.

pkg/controller/deorbit/events.go
```diff
@@ -16,6 +17,20 @@ type EventingDeorbiter struct {
 	Recorder  record.EventRecorder
 }
 
+func (d *EventingDeorbiter) DeleteSnapshots() (deleted []*unstructured.Unstructured, err error) {
+	deleted, err = d.Deorbiter.DeleteSnapshots()
+
+	for i, snapshot := range deleted {
+		if err == nil || i < (len(deleted)-1) {
```
Nuckal777, line 24: `i` is guaranteed to not exceed `len(deleted)` by the range expression. Also err needs to be only checked once outside of the loop.

Nuckal777 (changes requested):
Added my 2 cents.

pkg/controller/deorbit/events.go
Nuckal777: Can be simplified with an early return at `err != nil`.

pkg/controller/deorbit/deorbiter.go
uwe-mayer: done

pkg/controller/deorbit/events.go
uwe-mayer: Hopefully this easier to grasp:
It first logs all successfully deleted objects and the failed on last if exists.

PR #759, opened 2023-02-10
Add backup & audit labels to infoMetric

uwe-mayer:
Adds `Backup` and `Audit` values from `klusterSpec` to `infoMetric` to sort out `KubernikusKlusterLowOnObjectStoreQuota` where no `Audit` or `Backup` to Swift is enabled...

https://github.wdf.sap.corp/cc/ccloud-containers/issues/65

### DIFF
--- a/pkg/controller/metrics/kluster.go
+++ b/pkg/controller/metrics/kluster.go
@@ -66,7 +66,7 @@ func (collector *klusterCollector) Collect(ch chan<- prometheus.Metric) {
 			getCreatorFromAnnotations(kluster.Annotations),
 			getAccountFromLabels(kluster.Labels),
 			kluster.Spec.Backup,
-			*kluster.Spec.Audit,
+			getAuditFromSpec(kluster.Spec),
 		)
 	}
 
--- a/pkg/controller/metrics/kluster.go
+++ b/pkg/controller/metrics/kluster.go
@@ -29,7 +29,7 @@ func NewKlusterCollector(lister kubernikus_lister.KlusterLister) *klusterCollect
 		infoMetric: prometheus.NewDesc(
 			"kubernikus_kluster_info",
 			"Detailed information on a kluster",
-			[]string{"kluster_namespace", "kluster_name", "phase", "api_version", "chart_version", "chart_name", "creator", "project_id"},
+			[]string{"kluster_namespace", "kluster_name", "phase", "api_version", "chart_version", "chart_name", "creator", "project_id", "backup", "audit"},
 			nil,
 		),
 		lister: lister,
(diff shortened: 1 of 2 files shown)


pkg/controller/metrics/kluster.go
```diff
@@ -65,6 +65,8 @@ func (collector *klusterCollector) Collect(ch chan<- prometheus.Metric) {
 			kluster.Status.ChartName,
 			getCreatorFromAnnotations(kluster.Annotations),
 			getAccountFromLabels(kluster.Labels),
+			kluster.Spec.Backup,
+			*kluster.Spec.Audit,
```
databus23, line 69: Is this guaranteed to be non-nil all the time? Otherwise this would create a panic

uwe-mayer: As per typing it cannot (if my novice understanding of go is correct):
`invalid operation: cannot compare kluster.Spec.Backup == nil (mismatched types string and untyped nil)`
Did I miss a check?

See `kluster_spec.go`:
https://github.com/sapcc/kubernikus/blob/master/pkg/api/models/kluster_spec.go#L30-L36

databus23: `Backup` is fine because it not a pointer, but `Audit` is. dereferencing a pointer that is not set (`*pointername`) creates a panic crashing the process.

uwe-mayer: I was wondering why `Audit` was a pointer and Backup not. The `kluster_spec.go` is autogenerated by swagger right? So I guess this is the way of marking optional properties?

Nuckal777: Setting audit is optional, so null is valid value, which means do not audit log.https://github.com/sapcc/kubernikus/blob/master/swagger.yml#L472

pkg/controller/metrics/metrics.go
```diff
@@ -133,6 +133,13 @@ func getAccountFromLabels(labels map[string]string) string {
 	return account
 }
 
+func getAuditFromSpec(spec models.KlusterSpec) string {
+	if spec.Audit == nil {
+		return "NA"
```
databus23, line 138: I suggest to set it to the empty string `""` instead

uwe-mayer: I decided to go for "NA" for consistency:
https://github.com/sapcc/kubernikus/blob/master/pkg/controller/metrics/metrics.go#L120-L134

There is situations where the alertmanager evaluates an empty label matching a missing label:
https://prometheus.io/docs/alerting/latest/configuration/#inhibit_rule

uwe-mayer: I am happy to commit either one of them, you decide @databus23  :-)

databus23 (approved):

PR #776, opened 2023-03-16
Migrate kubedns to coredns

Nuckal777:
Since we reconcile content in customer clusters now, we can migrate people from kubedns to coredns with ease. 🥳


databus23 (approved):

databus23:
Looks good but want to ship the pending commits first before rolling this out in a staggered approach

PR #797, opened 2023-05-19
Add kluster specific user-agent to openstack api calls

databus23:


Nuckal777:
TODO: exclude old cluster versions in helm template.

PR #798, opened 2023-05-22
[WIP] Add support for Kubernetes 1.27

Nuckal777:
As this is from my local 1.27-rc branch, the semver regex was adjusted as well as the conditions in the helm chart, so that `1.27-rc... >= 1.27`.

- [ ] Test it
- [ ] Consider updating the CSI components


jknipper:
Closed because of #816

PR #806, opened 2023-06-13
Remove unschedulable nodes first on nodepool downsizing

jknipper:


pkg/controller/launch/pool_manager.go
```diff
@@ -303,3 +307,58 @@ func (cpm *ConcretePoolManager) healthyAndSchedulable() (healthy int, schedulabl
...
+		}
+	}
+
+	// sort abandoned openstack nodes to top
+	for _, oID := range openstackIDs {
+		pos := slices.Index(nodeIDs, oID)
+		if pos > 0 {
+			nodeIDs = toTop(nodeIDs, pos)
+		}
+	}
```
databus23, line 355: Pseudo code for sort function
```suggestion
        kubernetesIDs := map[string]Node
	sort.Slice(nodeIDs, func(i, y int) bool {
		if _, ok := kubernetesIDs[nodeIDs[i]]; !ok {
			return true
		}
		if _, ok := kubernetesIDs[nodeIDs[j]]; !ok {
			return false
		}
		if n, ok := kubernetesIDs[nodeIDs[i]]; ok && n.Spec.Unschedulable {
			return true
		}
		if n, ok := kubernetesIDs[nodeIDs[j]]; ok && n.Spec.Unschedulable {
			return false
		}
		return true
	})
```

databus23:
@jknipper I added some tests and made the sorting stable.

PR #821, opened 2023-07-13
Remove gpu support

jknipper:


Nuckal777:
I guess https://github.com/sapcc/kubernikus/tree/master/contrib/nvidia-gpu can be removed as well.

jknipper:
> I guess https://github.com/sapcc/kubernikus/tree/master/contrib/nvidia-gpu can be removed as well.

Good find, removed it as well.

databus23 (approved):

PR #834, opened 2023-10-09
fix(template): containerd plugin name

defo89:
* to fix error 
```
failed to load TOML from /etc/containerd/config.toml: invalid plugin key URI "containerd.runtime.v1.linux" expect io.containerd.x.vx
```


Nuckal777:
Will this crash on older flatcar nodes again? 🤔

databus23 (approved):

PR #848, opened 2023-11-22
Fix etcd and etcdbr images

jknipper:


databus23:
ups so many changes? Did we really miss that much or did you decide to back port the etcd upgrade to older k8s versions?
We had a discussion once about not upgrading older versions of etc

databus23 (approved):

databus23:
I see, makes sense! 👍

PR #850, opened 2023-12-06
Make torcx support optional

jknipper:
Torcx is going to be removed in future flatcar release, see https://www.flatcar.org/docs/latest/provisioning/sysext/#torcx-deprecation

- [x] Test upgrade by switching a stable vm to alpha channel

### DIFF
--- a/pkg/templates/node_1.26.go
+++ b/pkg/templates/node_1.26.go
@@ -34,7 +34,7 @@ systemd:
           contents: |
             [Service]
             ExecStart=
-            ExecStart=/usr/bin/env PATH=${TORCX_BINDIR}:${PATH} ${TORCX_BINDIR}/containerd
+            ExecStart=/usr/bin/containerd
     - name: docker.service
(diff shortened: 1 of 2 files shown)


Nuckal777:
Do we need to do something for old nodes upgrading into torcx removal, because they should not get the new script via ignition? 🧐

jknipper:
> Do we need to do something for old nodes upgrading into torcx removal, because they should not get the new script via ignition? 🧐

Good question, I have no idea if existing nodes will be migrated to `systemd-sysext` or not.

jknipper:
fyi @defo89

jknipper:
I tested switching a node to alpha release channel. It's not coming up anymore and is not joining the cluster. This means we need to prepare a fix for existing nodes.

defo89:
Just a thought for the upgrade case - if in both locations the file is not found, should we maybe copy it from same swift location (the region will need to be populated with some env though)?

PR #853, opened 2023-12-18
Disable reboots in servicing controller

jknipper:


databus23:
I'm not sure if this is the rough approach as this will not only disable os updates but also upgrades of nodes in case a user triggers an update.
Can we maybe just disable the flatcar os update?

jknipper:
Changed it to disable reboots only.

databus23 (approved):

PR #916, opened 2024-05-29
Seed virtual kluster

jknipper:
This PR introduces a new kluster setting `seedVirtual` which enables seeding of needed components for our virtual klusters. The setting is independent of `seedKubeadm`, don't enable both settings at the same time! It also removes some CSI stuff for both `seedKubeadm` and `seedVirtual` which is not needed in those klusters.

Todos:

- [x] testing
- [x] verify config

### DIFF
--- a/charts/seed/crds/volumesnapshot.yaml
+++ b/charts/seed/crds/volumesnapshot.yaml
@@ -1,3 +1,4 @@
+{{- if and (not .Values.seedKubeadm) (not .Values.seedVirtual) }}
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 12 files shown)


charts/seed/crds/volumesnapshot.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if and (not .Values.seedKubeadm) (not .Values.seedVirtual) }}
```
defo89, line 1: shouldn't this be an `or` condition - otherwise the ceph clusters will get this? same goes for all other resources with `and` condition

jknipper: I mean it's currently seeded in the ceph clusters. But isn't the condition `false` if either one or both are `true`? Also see the updated description.

PR #925, opened 2024-06-19
wormhole kubeconfig path when using seedVirtual

xsen84:


databus23 (approved):

SchwarzM (approved):
jan looked over it with me ... we hope it's good

PR #942, opened 2024-09-10
Fix old node templates

jknipper:


pkg/templates/ignition.go
```diff
@@ -238,7 +238,7 @@ func (i *ignition) GenerateNode(kluster *kubernikusv1.Kluster, pool *models.Node
 	)
 
 	if err != nil {
-		return nil, fmt.Errorf("Couldn't translate ignition file: %v", report.String())
+		return nil, fmt.Errorf("Couldn't translate ignition file: %v\n%v", report.String(), err)
```
Nuckal777, line 241: Nitpick: Errors should be concatenated with `%w`.

Nuckal777 (approved):
Thanks for the cleanup.

pkg/templates/ignition.go
jknipper: changed it

PR #952, opened 2024-11-05
Disable openstack client in csi nodeplugin

jknipper:
This disables the openstack client in the CSI nodeplugin and the need for a technical user secret in the cluster for k8s >= 1.31.2.

Todo:

- [ ] bump csi to upcoming version
- [ ] enable it in k8s 1.31.3


jknipper:
Looks like there is a bug in the 1.31 nodeplugin. It will be fixed in the next release.

PR #967, opened 2024-12-17
Update linter, fix loop references

jknipper:
As far as I understand behavior of references in loops is changed in golang>=1.22


Nuckal777:
I think our Dockerfile needs linter updates as well or the linter to be removed. 🤔

jknipper:
> I think our Dockerfile needs linter updates as well or the linter to be removed. 🤔

Which one do you mean? I only found it in the ones already included.

Nuckal777 (approved):
I mentally skipped the dockerfile adjustments. :(

PR #968, opened 2024-12-17
Disable OS updates during xmas

jknipper:


Nuckal777 (approved):

pkg/controller/servicing/reconciler.go
```diff
@@ -157,7 +158,8 @@ func (r *KlusterReconciler) Do() error {
 
 	update := r.Lister.Updating()
 	replace := r.Lister.Replace()
-	reboot := r.Lister.Reboot()
+	// reboot := r.Lister.Reboot()
+	reboot := make([]*core_v1.Node, 0) // disable reboots for now
```
SuperSandro2000, line 162: Wouldn't it be nicer to check that date? 🫣

PR #989, opened 2025-03-05
[flatcar] add oem-vmware.raw as an extra sysext

kayrus:
fixes https://github.com/flatcar/Flatcar/issues/1589


kayrus:
@jknipper 
* do we need to add an extra trigger to include this sysext?
* this fix is only for new nodes, do we need to cover existing nodes?

jknipper (approved):

jknipper (approved):

PR #994, opened 2025-03-27
Add admission config to apiserver

jknipper:


charts/kube-master/templates/configmap.yaml
```diff
@@ -255,3 +255,27 @@ data:
...
+    - name: MutatingAdmissionWebhook
+      configuration:
+        apiVersion: apiserver.config.k8s.io/v1
+        kind: WebhookAdmissionConfiguration
+        kubeConfigFile: "/etc/kubernetes/admission/kubeconfig"
+  admission-kubeconfig: |-
+    apiVersion: v1
+    kind: Config
+    users:
+    - name: '*'
```
databus23, line 277: I have never seen this. Does this work? 🤯

jknipper: I've seen this somewhere but can't recall where. I guess we need to try it out.

Nuckal777: [Here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#authenticate-apiservers)

Nuckal777 (approved):

jknipper:
Ok, disabled this by default for now.

PR #995, opened 2025-04-02
Initial support for structured-auth

databus23:
This is the most basic approach to add support for adding support for using structured authentication configuration instead of the `--oidc-*` flags.

I decided to keep the lifecycle management outside of the `kube-master` helm chart so that updates to the authentication configuration can be applied without upgrading the kubernetes version.

Haven't tested this.


Nuckal777:
Would it make sense to include the structured auth config type somewhere in the `Kluster` resource and apply a `ConfigMap` including that data with some reconciler? This way it could be made part of the API.

databus23:
> Would it make sense to include the structured auth config type somewhere in the Kluster resource and apply a ConfigMap including that data with some reconciler? This way it could be made part of the API.

I have thought about this and decided against it for the following reasons:

1. By adding it to the spec we expose this to kubernikus customers and allow them to use this and increase the support burden on our side. I wasn't sure we wanted to add this as a fully supported feature
2. If the configuration happens via the Kluster spec the effect of changes is delayed until the next kubernetes version upgrade because that is the only time where we apply the master chart. Rolling out any changes to the authentication configuration file is difficult, especially if you rolled out an incomplete change but have no kubernetes version upgrade left to  roll it out.

One upside of adding it to spec is that rolling this out for new cluster is easier. I think this is why we added support for the `--oidc` overrides to the spec in https://github.com/sapcc/kubernikus/pull/880 to smoothen the rollout of ceph clusters that are wrapped in a cluster-api provider atm

databus23:
@Nuckal777 I reworked the PR as per your suggestions to support setting the authentication configuration via an additional field in the cluster spec.

databus23:
@Nuckal777 kind ping. Any comments or concerns?

pkg/api/models/kluster_spec.go
```diff
@@ -187,6 +198,23 @@ func (m *KlusterSpec) validateAudit(formats strfmt.Registry) error {
 	return nil
 }
 
+func (m *KlusterSpec) validateAuthenticationConfiguration(formats strfmt.Registry) error {
```
Nuckal777, line 201: Would it make sense to check for the minimum kluster version here as well?

pkg/controller/ground.go
```diff
@@ -1187,3 +1205,79 @@ func (op *GroundControl) podUpdate(cur, old interface{}) {
...
+	var configContent string
+	if kluster.Spec.AuthenticationConfiguration != "" {
+		configContent = string(kluster.Spec.AuthenticationConfiguration)
+	} else {
+		authConfig := &apiserverv1beta1.AuthenticationConfiguration{}
+		gvks, _, err := apiserverScheme.ObjectKinds(authConfig)
+		if err != nil {
+			return fmt.Errorf("cannot determine GVK: %w", err)
+		}
+		authConfig.GetObjectKind().SetGroupVersionKind(gvks[0])
```
Nuckal777, line 1219: `apiserverv1beta1.SchemeGroupVersion.WithKind("AuthenticationConfiguration")` would be more concise.

pkg/controller/ground.go
```diff
@@ -1187,3 +1205,79 @@ func (op *GroundControl) podUpdate(cur, old interface{}) {
...
+	if existingConfigMap.Data["config.yaml"] != cm.Data["config.yaml"] {
+		_, err := op.Clients.Kubernetes.CoreV1().ConfigMaps(cm.GetNamespace()).Update(context.TODO(), cm, meta_v1.UpdateOptions{})
+		op.Logger.Log(
+			"msg", "Updating authentication configuration configmap",
+			"name", cm.GetName(),
+			"kluster", kluster.GetName(),
+			"err", err)
+		return err
+	}
+	return nil
```
Nuckal777, line 1281: With server-side apply this can potentially be shortened to a single api call.

pkg/controller/ground.go
```diff
@@ -1187,3 +1205,79 @@ func (op *GroundControl) podUpdate(cur, old interface{}) {
...
+
```
Nuckal777, line 1282: Superfluous newline.

Nuckal777 (commented):
Thanks. 👍  Maybe the validation in the update call kluster handler needs adjustments as well?

pkg/api/models/kluster_spec.go
databus23: That's hard to do. The kluster_spec.go is generated and I don't know of a way to add custom validation at this point.

pkg/controller/ground.go
databus23: I don't now how to prevent visual studio code from inserting this newline.
whenever I save its back. "Insert final new line" in code is already disabled.
Is this important?

pkg/controller/ground.go
databus23: I tried adding that's and while its a single api call it increases the lines of code so I don't see a big benefit in changes this. During runtime the current approach will only do a single `get` configmap call unless an update is necessary. Changing this to an `server side apply` call in every reconciliation round is probably more expensive for the apiserver.

pkg/controller/ground.go
databus23: thanks, updated

Nuckal777 (approved):

PR #1025, opened 2025-07-11
Allow to specific flavor and custom root disk in e2e test

databus23:
We want to run the kubernikus e2e test on KVM, for this we need a different flavor and custom root disks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ build-e2e:
 .PHONY: test-e2e
 test-e2e:
 ifdef KS_PASSWORD
-	export OS_PASSWORD=$(KS_PASSWORD)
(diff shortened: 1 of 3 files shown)


Nuckal777 (approved):

jknipper (approved):

databus23:
> I think the ssh key change requires adjustments in the pipeline.
Really? Leaving it empty means no ssh key will be provisioned which is the same as we have currently we we have no idea were the private key is for the static key we have in the tests. :)
I don't think we need the ssh key in our soak tests as `spore drill|ssh` provides a way to acces the nodes for debugging

Nuckal777 (approved):

PR #1027, opened 2025-07-29
Bump cloud provider openstack 1.31/1.32/1.33

kayrus:


databus23 (dismissed):

charts/images.yaml
```diff
@@ -7589,10 +7589,10 @@ imagesForVersion:
     wormhole:
       repository: keppel.global.cloud.sap/ccloud/kubernikus
       tag: changeme
-  1.31.3:
```
Nuckal777, line 7592: Why is the controlplane version changed here?

Nuckal777 (changes requested):

PR #1069, opened 2026-04-29
Add support for KVM

databus23:
For kvm we need to always use cinder based root disks and use a dedicated volume_type.

The basic idea is to enforce cinder based root disks and set the volume type when any of our lvm flavours is used.


jknipper (approved):

MaximShepelev (approved):
lgtm

PR #1072, opened 2026-05-21
chore: bump etcd to v3.4.36 and etcdbrctl to v0.41.2

s10:
Latest patch on the etcd 3.4 line, paired with the latest etcdbrctl release still built against the 3.4 client. Adds Dockerfile.etcdbrctl and Makefile targets.

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -8313,10 +8313,10 @@ imagesForVersion:
       tag: 38f4f8ea8d487470a1dd5b83d66b428d8b502f81
     etcd:
       repository: keppel.global.cloud.sap/ccloud/etcd
-      tag: v3.4.35
+      tag: v3.4.36
     etcdBackup:
       repository: keppel.global.cloud.sap/ccloud/etcdbrctl
-      tag: v0.25.1
+      tag: v0.41.2
     flannel:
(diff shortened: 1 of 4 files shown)


charts/images.yaml
```diff
@@ -8313,10 +8313,10 @@ imagesForVersion:
       tag: 38f4f8ea8d487470a1dd5b83d66b428d8b502f81
     etcd:
       repository: keppel.global.cloud.sap/ccloud/etcd
-      tag: v3.4.35
+      tag: v3.4.36
     etcdBackup:
       repository: keppel.global.cloud.sap/ccloud/etcdbrctl
-      tag: v0.25.1
+      tag: v0.41.2
```
kayrus, line 8319: this is a big gap with a previous version. despite minor version gap, there are some breaking changes in release notes. do we handle them?

s10: I haven't found any breaking changes that affect kubernikus's etcd/etcdbrctl deployment.
We don't use `initial-advertise-peer-urls`/`advertise-client-urls`, as well as `enable-etcd-member-gc`/`k8s-member-gc-duration`.

MaximShepelev (approved):

goerangudat (approved):
lgtm

PR #1073, opened 2026-05-26
fix(etcd): pin etcdbrctl --endpoints to match transport

s10:
etcd listens on the scheme picked by secure.enabled; etcdbrctl
expects --endpoints with the same scheme. Its default became http,
so without this the backup container can't reach an HTTPS etcd.


goerangudat (approved):
lgtm

MaximShepelev (approved):

PR #1074, opened 2026-05-27
fix(etcd): mount Swift credentials as files for etcdbrctl

s10:
etcdbrctl v0.28+ unconditionally checks Swift credential file mtime before every snapshot; without OPENSTACK_APPLICATION_CREDENTIALS set, it refuses to upload. Mount the etcd Secret as files at /var/etcd/swift-creds with upstream-expected names (tenantName, authURL, ...).

### DIFF
--- a/pkg/apis/kubernikus/v1/secret.go
+++ b/pkg/apis/kubernikus/v1/secret.go
@@ -69,6 +69,7 @@ type Openstack struct {
 	DomainName        string `json:"openstack-domain-name,omitempty"`
 	Password          string `json:"openstack-password"`
 	ProjectID         string `json:"openstack-project-id"`
+	ProjectName       string `json:"openstack-project-name,omitempty"`
 	ProjectDomainName string `json:"openstack-project-domain-name,omitempty"`
 	UserDomainID      string `json:"openstack-user-domain-id,omitempty"`
 	ProjectDomainID   string `json:"openstack-project-domain-id,omitempty"`
--- a/pkg/controller/ground.go
+++ b/pkg/controller/ground.go
@@ -680,6 +680,12 @@ func (op *GroundControl) createKluster(kluster *v1.Kluster) error {
 		}
 		klusterSecret.ProjectDomainName = domainNameByProject
 
+		projectName, err := adminClient.GetProjectName(klusterSecret.ProjectID)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve project name: %s", err)
+		}
+		klusterSecret.ProjectName = projectName
+
 		userDomainID, err := adminClient.GetDomainID(klusterSecret.DomainName)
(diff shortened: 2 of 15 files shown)


MaximShepelev:
Chart bump not needed [here](https://github.com/sapcc/kubernikus/blob/master/charts/kube-master/charts/etcd/Chart.yaml)?

pkg/apis/kubernikus/v1/secret.go
```diff
@@ -69,6 +69,7 @@ type Openstack struct {
 	DomainName        string `json:"openstack-domain-name,omitempty"`
 	Password          string `json:"openstack-password"`
 	ProjectID         string `json:"openstack-project-id"`
+	ProjectName       string `json:"openstack-project-name,omitempty"`
```
databus23, line 72: Do we really need to add the project name? doesn't work with the ID?

pkg/controller/ground.go
```diff
@@ -680,6 +680,12 @@ func (op *GroundControl) createKluster(kluster *v1.Kluster) error {
 		}
 		klusterSecret.ProjectDomainName = domainNameByProject
 
+		projectName, err := adminClient.GetProjectName(klusterSecret.ProjectID)
```
databus23, line 683: This code only covers new clusters. all existing clusters will not have the project-name set.
I would assume this will break when upgrading existing clusters and the swift creds will be written to disk without a project name

databus23: We have a migration concept in kubernikus to retrofit this value for existing clusters: https://github.com/sapcc/kubernikus/tree/master/pkg/migration

databus23 (changes requested):

pkg/apis/kubernikus/v1/secret.go
s10: etcd-backup-restore requires ProjectName, https://github.com/gardener/etcd-backup-restore/blame/master/pkg/snapstore/swift_snapstore.go#L69

It checks these files since https://github.com/gardener/etcd-backup-restore/commit/69581185#diff-c06db825652553231618a6ab062cbe6fccd179e30c76432af2a64a95e6431145R44-R519 and fails if any is absent

Then it expects tenant name, as it set OS_TENANT_NAME env https://github.com/gardener/etcd-backup-restore/blame/master/pkg/snapstore/swift_snapstore.go#L164

pkg/controller/ground.go
s10: Added migration 22.

s10:
@jknipper @databus23 , could you please look at this PR?

MaximShepelev (approved):

goerangudat (approved):
lgtm

databus23 (approved):

PR #1082, opened 2026-06-22
fix: revert etcd-backup-restore to 0.25.1

s10:
* guard check currently contains release name as a namespace variabe
* pod namespace is used in newer etcd-backup-restore versions


charts/kube-master/charts/etcd/templates/deployment.yaml
```diff
@@ -334,16 +334,14 @@ spec:
 {{- if (semverCompare ">= 1.23-0" .Values.version.kubernetes) }}
             - name: POD_NAME
 {{- if (semverCompare ">= 1.34-0" .Values.version.kubernetes) }}
-              value: "kubernikus"
+              value: {{ include "fullname" . }}
```
s10, line 337: Used in etcd.conf.yaml

PR #1083, opened 2026-06-23
chore(etcd): add init container to validate safe_guard namespace

jknipper:
Add an init container that checks if the etcd backup safe_guard file contains the same namespace as the current POD_NAMESPACE. If the namespace has changed (due to recent updates), the safe_guard file is removed to allow the backup restore sidecar to reinitialize.


stwilkes (approved):
lgtm

databus23 (approved):

